PR #11, opened 2016-12-12
bridge data provider constructor

dephillipsmichael:
Try to adjust bridge data provider so it can be used...  @liujoshua can you comment on if this is valid or not?


rianhouston:
Nice!  This is one of the issues I encountered.

PR #15, opened 2017-01-20
Feature/onboarding manager changes

dephillipsmichael:
Not meant to be merged, yet, since I switched grade to import a project.  However, this is meant to be reviewed, and used to get the MPower branch running. https://github.com/Sage-Bionetworks/mPower-Android/pull/19 and https://github.com/ResearchStack/ResearchStack/pull/320


researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -4,8 +4,10 @@
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 
+import com.fatboyindustrial.gsonjodatime.Converters;
```
liujoshua: Delete. I don't see this used here, and I plan to remove this dependency from rest-client.

researchstack-sdk/build.gradle
```diff
@@ -44,7 +44,7 @@ android {
 
 dependencies {
   compile fileTree(include: ['*.jar'], dir: 'libs')
-  compile 'org.researchstack:skin:1.1.1', {
+  compile project(':skin'), {
```
liujoshua, line 47: This reminds me that I need to find a build setup that works for both this repo itself, and when it's included as a submodule. Making a JIRA

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -16,23 +18,26 @@
 import org.researchstack.backbone.utils.LogExt;
 import org.researchstack.backbone.utils.ObservableUtils;
 import org.researchstack.skin.AppPrefs;
-import org.researchstack.skin.DataProvider;
-import org.researchstack.skin.DataResponse;
-import org.researchstack.skin.ResourceManager;
-import org.researchstack.skin.model.SchedulesAndTasksModel;
+import org.researchstack.backbone.DataProvider;
```
liujoshua, line 19: Do we have a high-level strategy or plan for how we want to split things between backbone, skin, and Bridge SDK?

I feel a more generic version of DataProvider could go in backbone, the current DataProvider seems pretty specific to Bridge... Or we can see how things evolve and clean up before merging into master.

liujoshua (commented):
Thanks

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
dephillipsmichael: I think we should see how things evolve and then do a clean up or re-organization once we move back into master.

As of right now, my thoughts for moving forward are that we should move everything from Skin into backbone when we can, and eventually delete Skin all together.  Backbone should have an abstract DataProvider, which encompasses every possible Data operation.  There should be a LocalDataProvider in backbone, which is a concrete implementation of DataProvider, that stores everything locally, and also is something for the developers to use as guidance on implementing  the app, or can be used if the app requires no internet access.  BridgeSDK would still have BridgeDataProvider, which would extend, or simply use LocalDataProvider, to also do all the server operations.

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
dephillipsmichael: good catch, deleted.

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
liujoshua: That sounds really great!

PR #16, opened 2017-01-23
Feature/authentication manager

liujoshua:
First pass at Auth Manager. To come:

- unit tests
- rename/refactor


android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/auth/AuthManager.java
```diff
@@ -0,0 +1,341 @@
...
+                                .password(password)
+                                .sptoken(resetToken)
+                ))
+                .doOnSuccess(new Action1<Message>() {
+                    @Override
+                    public void call(Message message) {
+                        authManagerDelegateProtocol.setPassword(password);
+                    }
+                }).toCompletable();
+    }
```
alxdarksage: I don't think you'll ever be in a position to call this. The request reset password call triggers an email and that has a link back to a web page on BridgePF where you can change the password. If this call is in the SDK, I'm not sure how we'd use it, even though this API is part of the Bridge API for that web page.

android-sdk/build.gradle
```diff
@@ -37,9 +38,16 @@ android {
 dependencies {
   compile fileTree(dir: 'libs', include: ['*.jar'])
 
+  compile 'com.madgag.spongycastle:core:1.54.0.0'
```
dephillipsmichael, line 46: What is spongy castle and how is it used within android-idk library?  I usually like to comment the gradle file for all 3rd-party libs.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/BridgeConfig.java
```diff
@@ -0,0 +1,153 @@
...
+
+        if (model.startsWith(manufacturer)) {
+            return capitalize(model);
+        } else {
+            return capitalize(manufacturer) + " " + model;
+        }
+    }
+
+    @NonNull
+    String capitalize(@Nullable String s) {
```
dephillipsmichael, line 142: This should probably live in backbone's TextUtils and be unit tested.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/auth/AuthManager.java
```diff
@@ -0,0 +1,341 @@
...
+        return RxUtils.toBodySingle(
+                authenticationApi.signIn(signIn))
+                .doOnSuccess(new Action1<UserSessionInfo>() {
+                    @Override
+                    public void call(UserSessionInfo userSessionInfo) {
+                        authManagerDelegateProtocol.setEmail(email);
+                        authManagerDelegateProtocol.setPassword(password);
+                        authManagerDelegateProtocol.setUserSessionInfo(userSessionInfo);
+                    }
+                }).doOnError(new Action1<Throwable>() {
```
dephillipsmichael: Should this be the place where we intercept the ConsentRequiredException?  Or will that be done by the AuthManagerDelegate, which I expect to be the BridgeDataProvider

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/auth/AuthManager.java
```diff
@@ -0,0 +1,341 @@
...
+     *
+     * @return notifies of completion or error
+     */
+    @NonNull
+    public Completable signOut() {
+        return RxUtils.toBodySingle(authenticationApi.signOut())
+                .doOnSuccess(new Action1<Message>() {
+                    @Override
+                    public void call(Message message) {
+                        authManagerDelegateProtocol.setEmail(null);
```
dephillipsmichael: Should we also clear the consent data here?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/util/retrofit/BodyOnSubscribe.java
```diff
@@ -0,0 +1,88 @@
+/*
+ * Copyright (C) 2016 Jake Wharton
```
dephillipsmichael, line 2: didn't he write retrofit?  seems odd code like this wouldn't just be in the library

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -50,7 +51,9 @@
 * This is a very simple implementation that hits only part of the Sage Bridge REST API
 * a complete port of the Sage Bridge Java SDK for android: https://github.com/Sage-Bionetworks/BridgeJavaSDK
  */
-public abstract class BridgeDataProvider extends DataProvider {
+@AnyThread
+@Deprecated
```
dephillipsmichael: Good idea.   The way were going I feel like we are going to have to release RS as RS2 😃

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider2.java
```diff
@@ -0,0 +1,202 @@
...
+
+    @Override
+    @Nullable
+    @Deprecated
+    public String getUserSharingScope(Context context) {
+        return getUserSharingScope();
+    }
+
+    @Nullable
+    public String getUserSharingScope() {
```
dephillipsmichael: Should this be an Enum?

dephillipsmichael:
Looks pretty good to me.  Some comments overall, but I know this is a work in progress.

android-sdk/build.gradle
liujoshua: It is used for encryption of Bridge data for uploads. I moved it from researchstack-sdk to android-sdk, since we'll want it in the general Android SDK

android-sdk/src/main/java/org/sagebionetworks/bridge/android/BridgeConfig.java
liujoshua: Good point

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/auth/AuthManager.java
liujoshua: Internally, the Managers might intercept ConsentRequiredException and do something in response, but I'm thinking I'm thinking all APIs which throw ConsentRequiredException would propagate it up to the app. One general use case is if the study managers publish a new required consent, and I think having an exception would be the best way to handle it since it  likely means whatever API you just called failed, and will fail if repeated, so the next time the app is in the foreground, the user should be prompted for consent.

I'm imagining this would mean we could have a global way of handling ConsentRequireException, including sign in. If that doesn't work out, this would be a good place to intercept it, but I think this would mean the app should make another call to check consent status.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/auth/AuthManager.java
liujoshua: Yes, I'll add a TODO. I'm not sure yet how trigger ConsentManager to clean up the user's data and we probably want other Managers to clear out other data, like Tasks/Activities in the DB, maybe bridge data pending uploads (but maybe not)

android-sdk/src/main/java/org/sagebionetworks/bridge/android/util/retrofit/BodyOnSubscribe.java
liujoshua: It's in the retrofit-adapters library but package-private. I'm stealing the code that lets you write a Retrofit service with Single, Completable, Observable, etc return types instead of Call'

We're using an auto-generated client, and currently don't know how much work it'd be to get it to generate these new responses and get handled by the adapter layer

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
liujoshua: I've basically given up on maintaining backwards compatibility :)

I did this to have a clean BridgeDataProvider file and intend that we replace the old one rather than have one extend the other

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider2.java
liujoshua: Yes, thanks.

.travis.yml
```diff
@@ -8,8 +8,12 @@ android:
   components:
     - platform-tools
     - tools
-    - build-tools-25.0.0
+    - build-tools-25.0.2
     - android-25
+
+    # RS components
+    - build-tools-23.02
+    - android-23
```
dephillipsmichael: No longer needed.  RS should always be the same as BASdk, so I'd just go ahead and update RS to 25.0.2.  It should be at 25 currently.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/BridgeManagerProvider.java
```diff
@@ -0,0 +1,52 @@
...
+        checkNotNull(config);
+
+        this.config = config;
+        this.authenticationManager = new AuthenticationManager(config);
+        this.studyParticipantManager = new StudyParticipantManager(authenticationManager);
+    }
+
+    @NonNull
+    public AuthenticationManager getAuthenticationManager() {
+        return this.authenticationManager;
```
dephillipsmichael, line 45: super minor: "this." not needed

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/StudyParticipantManager.java
```diff
@@ -0,0 +1,45 @@
...
+import rx.functions.Action1;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Created by jyliu on 1/20/2017.
+ */
+@AnyThread
+public class StudyParticipantManager {
+    @NonNull
```
dephillipsmichael, line 21: I like your proper use of @NonNull I need to do that more as well

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/StudyParticipantManager.java
```diff
@@ -0,0 +1,45 @@
...
+    private final AuthenticationManager authenticationManager;
+
+    public StudyParticipantManager(@NonNull AuthenticationManager authenticationManager) {
+        checkNotNull(authenticationManager);
+
+        this.authenticationManager = authenticationManager;
+    }
+
+    @NonNull
+    public Completable updateParticipant(@NonNull StudyParticipant studyParticipant) {
```
dephillipsmichael, line 31: Method header?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/auth/AuthenticationManager.java
```diff
@@ -0,0 +1,303 @@
...
+    @NonNull
+    private final DAO dao;
+    @NonNull
+    private final BridgeConfig config;
+    private final AuthenticationApi authenticationApi;
+    private ApiClientProvider apiClientProvider;
+    private ProxiedForConsentedUsersApi proxiedForConsentedUsersApi;
+
+    public AuthenticationManager(@NonNull BridgeConfig config) {
+        this(config, new InMemoryDAO(config.getApplicationContext()));
```
dephillipsmichael, line 43: is InMemoryDAO an Android class?  I've never heard of it.  Also, maybe a TODO: here saying this is not secure, must use keystore

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/auth/ProxiedForConsentedUsersApi.java
```diff
@@ -0,0 +1,175 @@
...
+
+import java.util.List;
+
+import retrofit2.Call;
+import retrofit2.http.Body;
+import retrofit2.http.Path;
+import retrofit2.http.Query;
+
+/**
+ * Proxies ForConsentedUserApi calls so it uses retrieves authenticated client from AuthenticationManager
```
dephillipsmichael, line 32: "uses retrieves authenticated" typo?

dephillipsmichael: This class' existence seems unnecessary to me, can you explain it more?

android-sdk/src/test/java/org/sagebionetworks/bridge/sdk/android/ExampleUnitTest.java
```diff
@@ -10,6 +19,9 @@
 public class ExampleUnitTest {
   @Test
   public void addition_isCorrect() throws Exception {
+
+    Logger log = LoggerFactory.getLogger(ExampleUnitTest.class);
```
dephillipsmichael, line 23: Does this mean Log statements using LoggerFactory work when called from unit tests now?

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider2.java
```diff
@@ -0,0 +1,285 @@
...
+                                           @NonNull String username, @NonNull String password) {
+        return signUp(email, password);
+    }
+
+    @NonNull
+    public Observable<DataResponse> signUp(@NonNull String email, @NonNull String password) {
+        checkNotNull(email);
+        checkNotNull(password);
+
+        return bridgeManagerProvider.getAuthenticationManager()
```
dephillipsmichael: move bridgeManagerProvider.getAuthenticationManager() to helper method called getAuthManager()?

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider2.java
```diff
@@ -0,0 +1,285 @@
...
+    @Override
+    @Nullable
+    @Deprecated
+    public User getUser(Context context) {
+        return getUser();
+    }
+
+    @Override
+    public void setUser(Context context, User user) {
+        throw new UnsupportedOperationException();
```
dephillipsmichael, line 89: Blocker... Why is this Unsupported?

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider2.java
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    @Nullable
+    public User getUser() {
+        AuthenticationManager.DAO DAO =
+                bridgeManagerProvider.getAuthenticationManager()
+                        .getDao();
+
+        User user = new User();
+        user.setEmail(DAO.getEmail());
```
dephillipsmichael: What about user's name, birthdate, sharing scope?

dephillipsmichael:
needs updated to point at RS develop

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/auth/ProxiedForConsentedUsersApi.java
liujoshua: We get an instance of ForConsentedUsersApi from a factory in Retrofit, and this instance is tied to a sign in (email, password pair). This proxy will always use the latest sign in and transparently delegating to a new instance of ForConsentedUsersApi if the password changes.

So any instance of ProxiedForConsentedUsersApi will automatically get password changes known to AuthManager, while the base ForConsentedUsersApi will throw exceptions.

Now that we're talking about it, this does seem kludgy. I think a change to the way we're tying credentials to a retrofit instance might be cleaner.

android-sdk/src/test/java/org/sagebionetworks/bridge/sdk/android/ExampleUnitTest.java
liujoshua: I couldn't figure out how to configure logback to log to console, but by also adding simple-logger (which you're not supposed to do, since you're supposed to provide one unique impl) it works for me.

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider2.java
liujoshua: I'm thinking those will go in ConsentManager

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/auth/ProxiedForConsentedUsersApi.java
dephillipsmichael: I think so too.  Seems there should be some way to inject the password into the ForConsentedUsersApi that Retrofit creates

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider2.java
dephillipsmichael: Not a blocker, ignore this

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider2.java
dephillipsmichael: I think there should be one access point for a User object.  Where that is, doesn't really matter, but it will create confusion otherwise when this class only returns a User with their email set, instead of all the user fields.

liujoshua:
I couldn't update to the newest from develop, had issues once the abortOnLint=false was removed. I picked a merge from 2 days ago so it shouldn't be horribly out of date.

Can update from develop again once this PR is merged
https://github.com/ResearchStack/ResearchStack/pull/338

dephillipsmichael (commented):
I'm ok with you merging this once it's been tested with MPower

dephillipsmichael (approved):

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/auth/ProxiedForConsentedUsersApi.java
EricSiegNW: Is retrofit actually necessary? While its an awesome library, are we creating more work for ourselves by using it?

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider2.java
```diff
@@ -0,0 +1,284 @@
...
+import rx.functions.Action0;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static org.sagebionetworks.bridge.researchstack.ApiUtils.SUCCESS_DATA_RESPONSE;
+
+/**
+ * Created by jyliu on 1/20/2017.
+ */
+@AnyThread
+public class BridgeDataProvider2 extends DataProvider {
```
EricSiegNW, line 33: This name seems concerning, and smells of us having more classes/stuff than we actually should. Can you summarize the structure?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/BridgeManagerProvider.java
```diff
@@ -0,0 +1,52 @@
...
+import org.sagebionetworks.bridge.android.BridgeConfig;
+import org.sagebionetworks.bridge.android.manager.auth.AuthenticationManager;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Created by jyliu on 1/20/2017.
+ */
+@AnyThread
+public class BridgeManagerProvider {
```
EricSiegNW, line 16: I'm confused as to why this class exists

rianhouston:
I checked out this branch and then tried to follow the classes involved in making an auth call from mPower.  It looks something like this: 

DataProvider -> MPowerDataProvider -> BridgeDataProvider -> BridgeDataProvider2 -> BridgeManagerProvider -> AuthenticationManager -> ApiClientManger -> AuthenticationAPI

Maybe this is still a work in progress and some of these will be going away?  Can we simplify this?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/auth/InMemoryDAO.java
```diff
@@ -0,0 +1,77 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Created by jyliu on 1/20/2017.
+ */
+@AnyThread
+class InMemoryDAO implements AuthenticationManager.DAO {
```
rianhouston, line 18: Is this class meant for unit testing?  I assume we will have another implementation for the app itself?

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider2.java
liujoshua: This is temporary.  I want to avoid blocking progress by breaking authentication so I am building this along side BridgeDataProvider.  This code is a work in progress and only simple happy cases have been tested.

My plan is for BridgeDataProvider (old code) to temporarily extend BridgeDataProvider2 (new code). This is to make it easier to run a mix of old and new code, e.g. commenting out methods in BridgeDataProvider, to allow BridgeDataProvider2's new code to be called, or in the old code, super.methodName() can be called so both old and new code are invoked (useful for things that should change state in both, like if the credentials are changed)

liujoshua: Unsure why I can't directly reply to your comment about Retrofit.

Retrofit comes from rest-client, which is Bridge's java SDK shared between Android and service code. It's generated via swagger, so theoretically we could switch away from Retrofit, but it would take some doing since a significant amount of Retrofit and OkHttp3 functionality has been built into rest-client.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/BridgeManagerProvider.java
liujoshua: This class is intended be a single point for configuration and set up of dependencies and Managers. Pass in the context once on application start, and BridgeManagerProvider can be called from any thread/context (background service, foreground activity) to get a Manager.

I think it'll make more sense once I add in more code. One benefit of having this centralized source for interacting with Bridge is it makes it easy to inject a dependency, such as a mock for testing purposes. Code from anywhere can call BridgeManagerProvider.getInstance().getXYZManager, and get the mock.

This is a work in progress too, I've made some minor changes in an upcoming PR.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/auth/InMemoryDAO.java
liujoshua: Yes, this is a placeholder. There will be another implementation. Michael is working on how to store credentials and I'll integrate with that.

(This is another reason I created BridgeDataProvider2, AuthManager doesn't have sufficient features to replace the old code)

PR #17, opened 2017-01-23
Reenable deploy script inclusion on master

liujoshua:
This will allow master branch to deploy artifact to binTray.

develop will still have these scripts commented out to play nicely with mPower-Android, which is including this repo as a submodule
develop will not deploy due to .travis.yml


liujoshua:
Usually, git-flow calls for a cascading merge to develop after merging a branch to master, but in this case it'd be better to wait until the script inclusion issue is fixed

PR #18, opened 2017-01-24
Feature/build latest researchstack

liujoshua:
Might need to add build-tools for specific versions

### DIFF
--- a/.travis/setup.sh
+++ b/.travis/setup.sh
@@ -1,0 +1,7 @@
+# hack to get latest ResearchStack code since artficts aren't being published
+if [[ $TRAVIS_BRANCH != 'master' ]]; then
+  git clone -b develop https://github.com/ResearchStack/ResearchStack.git
+  pushd ResearchStack
+  ./gradlew install
+  popd
+fi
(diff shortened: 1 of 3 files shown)


liujoshua:
Commit d2b6a11 should succeed, commit a048405 is expected to fail until this PR is merged https://github.com/ResearchStack/ResearchStack/pull/323

liujoshua:
Note that this will mean this repo's `develop` builds are no longer reproducible, since `skin:2.0.0-SNAPSHOT` will change with time

.travis/setup.sh
```diff
@@ -0,0 +1,8 @@
+# hack to get latest ResearchStack code since artficts aren't being published
+if [[ $TRAVIS_BRANCH != 'master' ]]; then
```
dephillipsmichael, line 2: would "$TRAVIS_BRANCH == 'develop'" make more sense?

dephillipsmichael:
"Note that this will mean this repo's develop builds are no longer reproducible, since skin:2.0.0-SNAPSHOT will change with time" hmmm not sure if that is the behavior we want or not.  What do you think about including RS as a sub-module?  That way we can control what that sub-module points to

liujoshua:
We could do that temporarily and change mPower's structure to mPower -> BridgeAndroidSDK -> ResearchStack

PR #20, opened 2017-01-26
Submodule

dephillipsmichael:
Switched BridgeAndroidSdk over to using a sub-module of ResearchStack.

The main motivation for this was that we could not get Eric's machine working with local maven deployment even though we followed Josh's instructions that worked for me.  These changes should allow any developer to integrate BridgeAndroidSdk and it should just work.

MPower PR https://github.com/Sage-Bionetworks/mPower-Android/pull/20 depends on this


build.gradle
```diff
@@ -32,32 +33,38 @@ buildscript {
...
-      source = android.sourceSets.main.java.srcDirs
-      classpath += project.files(android.getBootClasspath().join(File.pathSeparator))
-      classpath += fileTree(dir: "$buildDir/intermediates/exploded-aar/", include: "**/classes.jar")
-      classpath += configurations.compile
-      // Exclude generated files
-      exclude '**/BuildConfig.java'
-      exclude '**/R.java'
-    }
+    // These two projects are JARed and DOCed using their own build.gradle files
+    if (!project.projectDir.absolutePath.contains("backbone") && !project.projectDir.absolutePath.contains("skin")) {
```
dephillipsmichael, line 38: This seems like I changed a lot, but I only add the conditional, the rest is just indented.

dephillipsmichael: Also, @liujoshua what is the purpose of creating the JAR task?  My guess  would be to deploy it to bintray, but I wasn't completely sure

.gitignore
```diff
@@ -21,35 +11,66 @@
 build/
 gen/
 
-# Gradle files
-.gradle/
-build/
+#Exclude .pem files to make sure they aren't checked in
+*.pem
```
liujoshua, line 15: Definitely safe and generally true, so leave it for now.

I will be checking in the pem for our demo study at some point though. We want people to be able to check out our code and play with it without doing a bunch of setup. So BridgeAndroidSDK and BridgeResearchStackSampleApp

.gitignore
```diff
@@ -21,35 +11,66 @@
...
+
+# External tool builders
+*/.externalToolBuilders/
+
+# Locally stored "Eclipse launch configurations"
+*.launch
+
+# Android Studio local workspace
+workspace.xml
+tasks.xml
```
liujoshua, line 48: I would think with the nesting of projects we'd want to keep the **/.idea

build.gradle
liujoshua: We deploy to bintray and ask bintray to forward on to jcenter (which is nice because jcenter is a default artifact repo for Android Studio). 

Bintray requires source jars and javadocs to enable maven magic that also helps us as devs (probably not needed now that we're using submodules) which is having access to published source code and javadocs instead of decompiled code in the IDE.

liujoshua: I like the idea of submodules, but needing to do things like this makes me wonder if we're not using gradle correctly. I briefly looked into this in December but couldn't find examples of how to do nested builds that should also work as standalone builds. It imagine it's possible because what we're trying to do with these submodules is reasonable and pretty common

I think the reason this happens is we're essentially ignoring ResearchStack's top level build.gradle (which is bad for many reasons). And likewise, mPower is ignore this build.gradle and also changing the pathing, which is why we had the issue of  not being able to find artifact-deploy script. When we enter a submodule, it'd expect gradle to pick up that submodule's top level build script and use it to build that submodule

liujoshua (approved):
Ship it!

build.gradle
dephillipsmichael: Why is it bad to ignore RS' top level gradle?  I don't see anything important in it, and if there is, we should move them into backbone's build.gradle or skin's.  Plus, that is just how gradle works, the top-level is only applicable for running the top-level project.

PR #23, opened 2017-02-03
Feature/data archive

liujoshua:
Wrote code for the steps from creating an archive to finishing the upload.

TODOs
- tie the pieces together
- persist upload sessions and status
- test encryption


android-sdk/src/main/java/org/sagebionetworks/bridge/android/BridgeConfig.java
```diff
@@ -98,6 +98,11 @@ public int getAppVersion() {
     }
 
     @NonNull
+    public String getAppVersionName() {
+        return applicationContext.getResources().getString(R.string.osb_app_version_name);
```
dephillipsmichael: Why wouldn't we use BuildConfig's versionName or versionCode for this?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/data/Archive.java
```diff
@@ -0,0 +1,121 @@
...
+
+        ZipEntry infoFileEntry = new ZipEntry(ARCHIVE_INFO_FILE_NAME);
+        zos.putNextEntry(infoFileEntry);
+        zos.write(RestUtils.GSON.toJson(archiveInfo).getBytes());
+        zos.closeEntry();
+        zos.close();
+        return zos;
+    }
+
+    public static class Builder implements WithBridgeConfig {
```
dephillipsmichael, line 53: Bonus points for Builder pattern

android-sdk/src/main/java/org/sagebionetworks/bridge/android/data/ArchiveInfo.java
```diff
@@ -0,0 +1,93 @@
...
+import org.joda.time.DateTime;
+
+import java.util.List;
+
+import static com.google.common.base.Strings.isNullOrEmpty;
+
+/**
+ * This is the model for info.json
+ */
+@SuppressWarnings("unused")
```
dephillipsmichael, line 16: This is used?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/data/StudyUploadEncryptor.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+public class StudyUploadEncryptor {
+    private static final Logger LOG = LoggerFactory.getLogger(StudyUploadEncryptor.class);
+
+    private final Supplier<JceKeyTransRecipientInfoGenerator> recipientInfoGeneratorSupplier;
+
+    public StudyUploadEncryptor(X509Certificate publicKey) {
+        this.recipientInfoGeneratorSupplier = Suppliers.memoize(() -> {
+            try {
+                return new JceKeyTransRecipientInfoGenerator(publicKey).setProvider("SC");
```
dephillipsmichael: What is "SC"?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/BridgeConfig.java
liujoshua: Good point. For Bridge, version name needs to be in a specific format, but I don't see a problem with documenting that and using the standard versionCode and versionName

android-sdk/src/main/java/org/sagebionetworks/bridge/android/data/Archive.java
liujoshua: :) In this case it makes it harder to use incorrectly, e.g. set the fields for surveys and non-survey activities and make an ambiguous data file

android-sdk/src/main/java/org/sagebionetworks/bridge/android/data/ArchiveInfo.java
liujoshua: This class is used for GSON which uses reflection to access the fields, so this suppresses IntelliJ's warnings about fields that are never set or retrieved.

Another option is to tell IntelliJ to ignore for fields that are annotated with `@SerializedName`. I think this setting can be done project level and shared.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/data/StudyUploadEncryptor.java
liujoshua: Spongy Castle, the encryption provider we're depending on. (BouncyCastle patched for Android)
I'll document this and make this a constant

android-sdk/src/main/java/org/sagebionetworks/bridge/android/data/StudyUploadEncryptor.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * @param stream plaintext stream
+     * @return encrypted stream
+     * @throws CMSException problem with encryption
+     * @throws IOException problem with stream
+     */
+    public OutputStream encrypt(OutputStream stream) throws CMSException, IOException {
+        JceKeyTransRecipientInfoGenerator recipientInfoGenerator = recipientInfoGeneratorSupplier
```
dephillipsmichael, line 50: Using Spongy castle here just increases our dependency on it, when eventually we want move away from 3rd party libraries.  Using the X509 public key to encrypt the stream should be fairly straightforward with the Cipher class, did you take a look at using that already?  I'd consider this a blocker.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
```diff
@@ -0,0 +1,143 @@
...
+
+        DigestOutputStream md5OutStream = new DigestOutputStream(sink.openBufferedStream(), md5);
+
+        OutputStream encryptedOutputStream = encryptor.encrypt(md5OutStream);
+
+        archive.writeTo(encryptedOutputStream);
+
+        String md5Hash = BaseEncoding.base64().encode(md5.digest());
+
+        //TODO: write file metadata to db
```
dephillipsmichael, line 133: We could just write the raw file with the EncryptionProvider in StorageAccess.  Is that what you were thinking?

android-sdk/src/test/java/org/sagebionetworks/bridge/android/data/ByteSourceArchiveFileTest.java
```diff
@@ -0,0 +1,28 @@
...
+public class ByteSourceArchiveFileTest {
+    private static final String FILENAME = "archive-file-1.json";
+    private static final DateTime END_DATE = DateTime.now();
+    private static final ByteSource BYTE_SOURCE = ByteSource.wrap(new byte[]{1, 2, 3});
+
+    @Test
+    public void testCreationFromByteSource() throws IOException {
+        ByteSourceArchiveFile archiveFile = new ByteSourceArchiveFile(FILENAME, END_DATE, BYTE_SOURCE);
+
+        assertEquals(FILENAME, archiveFile.getFilename());
```
dephillipsmichael, line 24: Great test ;)

android-sdk/src/main/java/org/sagebionetworks/bridge/android/data/StudyUploadEncryptor.java
dephillipsmichael: Not a blocker, looked into it deeper.

dephillipsmichael (approved):
Minor comments, fix before merging

android-sdk/src/main/java/org/sagebionetworks/bridge/android/BridgeConfig.java
liujoshua: @DwayneJengSage do you know the format of AppVersionName expected in a Bridge data archive's info.json file?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
liujoshua: My next PR will store things in SharedPreferences, that's for android-sdk.

That is a good point though, the researchstack-sdk should probably use StorageAccess instead of SharedPreferences directly. I'll include that.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/BridgeConfig.java
DwayneJengSage: It used to be in the form "version 1.0.8, build 27", where the version was what matched the app store version and the build number was what matched the user_agent header. However, we've stopped enforcing (or even parsing) this. I'd still recommend keeping that format (or one similar to it), since it includes both the researcher-facing version number and the dev-facing version number.

(If you're wondering why there's more than one version number, that's a good question. I assume it comes from the days of iOS development, and I'm not sure it was "designed" as much as it "came into existence".)

liujoshua: If the header is "StudyName/X (Nexus 6P; Android 23) BridgeSDK/Y", then versionName is recommended to be "version A.B.C, build X"

Does that sound right?

DwayneJengSage: That is correct.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/data/ArchiveInfo.java
```diff
@@ -0,0 +1,93 @@
...
+    String appVersion;
+    @SerializedName("phoneInfo")
+    String phoneInfo;
+
+    @SerializedName("files")
+    List<FileInfo> files;
+
+    // used for surveys
+    @SerializedName("SurveyGuid")
+    String surveyGuid;
```
DwayneJengSage, line 42: Should be surveyGuid? (With lowercase s?)

It would definitely be worth writing a unit test that serializes an ArchiveInfo into JSON, and then validating the JSON looks the way we expect it to. This would catch the aforementioned bug, and would also help validate our code in general.

android-sdk/src/test/java/org/sagebionetworks/bridge/android/data/ArchiveBuilderTest.java
```diff
@@ -0,0 +1,126 @@
...
+        assertNotNull(bytes2);
+        assertArrayEquals(BYTES, bytes2.toByteArray());
+
+        ByteArrayOutputStream bytesInfo = map.get("info.json");
+        assertNotNull(bytesInfo);
+
+        ArchiveInfo info = RestUtils.GSON.fromJson(bytesInfo.toString(), ArchiveInfo.class);
+
+        assertEquals(APP_VERSION_NAME, info.appVersion);
+        assertEquals(DEVICE_NAME, info.phoneInfo);
```
DwayneJengSage, line 111: Would it be worthwhile validating surveyGuid and surveyCreatedOn? What about a separate test validating item and schemaRevision?

PR #27, opened 2017-02-09
Feature/integrate managers

liujoshua:


liujoshua:
This PR supercedes the last one introducing the non-auth managers: https://github.com/Sage-Bionetworks/BridgeAndroidSDK/pull/26

android-sdk/src/main/java/org/sagebionetworks/bridge/android/BridgeApplication.java
```diff
@@ -17,5 +18,6 @@
     public void onCreate() {
         super.onCreate();
         JodaTimeAndroid.init(this);
+        BridgeManagerProvider.init(this);
```
dephillipsmichael, line 21: Nice idea

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AccountDAO.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.sagebionetworks.bridge.rest.model.SignIn;
+import org.sagebionetworks.bridge.rest.model.StudyParticipant;
+import org.sagebionetworks.bridge.rest.model.UserSessionInfo;
+
+/**
+ * Created by jyliu on 2/8/2017.
+ */
+@AnyThread
+public interface AccountDAO {
```
dephillipsmichael: What's the benefit of abstracting this class as an interface?  Seems confusing to me.  How about just making it a class and deleting InMemoryDao?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
```diff
@@ -269,35 +251,26 @@ public UserSessionInfo getUserSessionInfo() {
         // TODO: a way to distinguish if session is null because we haven't signed on, or if it
         // was invalidated
         UserSessionInfo session = apiClientProvider.getUserSessionInfoProvider()
```
dephillipsmichael, line 259: I don't think it is a good idea to have the UserSession exists in multiple places.  It leads to confusion about when you should use one or the other.  Is there a way we could just store it in the accountDao?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
```diff
@@ -269,35 +251,26 @@ public UserSessionInfo getUserSessionInfo() {
...
-        public void onLowMemory() {
-
-        }
+    /**
+     * Call Bridge to get session. Cached session is updated as a side-effect.
+     *
+     * @return Bridge session
+     */
+    @NonNull
+    public Single<UserSessionInfo> getLatestUserSessionInfo() {
```
dephillipsmichael, line 276: Should this be protected?  When would this be used?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
liujoshua: My intention is to expose two ways to get the UserSessionInfo, one for synchronous use (returns from accountDao only) and one async which may make a network call (and has to be done on the background thread). 

The devs should be using the Manager, and shouldn't have a way of getting at apiClientProvider or need to know  details of how UserSessionInfo is handled in rest-client. The only people who need to know how all that works are people contributing to rest-client and android-sdk.

I definitely want to expose only one way (actually two, sync and async) to get UserSessionInfo, StudyParticipant, Consents, so let me know if I missed anything.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
liujoshua: I made a comment earlier about wanting to provide a sync and async method. The async one here would be when you are looking for a new Session, not from cache. 

However, the cached Session should already remain up-to-date with respect to overall consent state (which seems to be most important), since a 412 will also cause the session in the cache to get updated. 

I'll look into making it protected, there might not be a need after all

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
liujoshua: For more context for the curious, the UserSessionInfoProvider provides sessions for multiple signIns, and is more general than what the AuthenticationManager does. Currently, all the managers are synced to one signIn (registered upon calling authManager.signIn).

UserSessionInfoProvider itself is backed by a class that intercepts HTTP requests to extract UserSessionInfo from 412s, calls which return UserSessionInfo, and track signIn/signOut.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/BridgeManagerProvider.java
```diff
@@ -1,52 +1,151 @@
...
+    public static void init(@NonNull Context applicationContext) {
+        checkNotNull(applicationContext);
+        checkState(instance == null, "BridgeManagerProvider has already been initialized");
+
+        instance = new BridgeManagerProvider(applicationContext);
+    }
+
+    private BridgeManagerProvider(@NonNull Context applicationContext) {
+        this.applicationContext = applicationContext;
+        bridgeConfig = new BridgeConfig(this.applicationContext);
```
dephillipsmichael, line 63: I'd suggest ditching the memoize call, IMO it's unnecessary confusion. The fact that this method is private and has a check state before makes it safe that these will only be called once.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ConsentDAO.java
```diff
@@ -0,0 +1,24 @@
...
+import android.support.annotation.Nullable;
+
+import org.sagebionetworks.bridge.rest.model.ConsentSignature;
+
+import java.util.Set;
+
+/**
+ * Created by jyliu on 2/8/2017.
+ */
+public interface ConsentDAO {
```
dephillipsmichael: Same goes for this, probably should just remove the abstraction

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ConsentManager.java
```diff
@@ -0,0 +1,268 @@
...
+        if (subpopulationStatus != null && subpopulationStatus.getConsented()) {
+            return true;
+        }
+        return dao.get(subpopulationGuid) != null;
+    }
+
+    /**
+     * @return true if all required consents have been signed
+     */
+    public boolean isConsented() {
```
dephillipsmichael, line 98: There are 3 public methods that start with "isConsented".  Can we make some private or protected so that it is clear which one returns the true status of if the user is consented?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ConsentManager.java
```diff
@@ -0,0 +1,268 @@
...
+     * @param sharingScope      participant's sharing scope for the study
+     * @return completable
+     */
+    @NonNull
+    public Completable giveConsent(@NonNull String subpopulationGuid, @NonNull String name,
+                                   @NonNull LocalDate birthdate,
+                                   @NonNull String base64Image, @NonNull String imageMimeType,
+                                   @Nullable SharingScope sharingScope) {
+        checkNotNull(subpopulationGuid);
+        checkNotNull(name);
```
dephillipsmichael: Some consent processes do not require name or birthdate, so I'm not sure whether we should checkNotNull for them.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ConsentManager.java
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    /**
+     * Sends a copy of the participant's signed consent to their email address.
+     *
+     * @param subpopulationGuid guid for the subpopulation of the consent
+     * @return completable
+     */
+    @NonNull
+    public Completable emailConsent(@NonNull String subpopulationGuid) {
```
dephillipsmichael: Isn't this done by the server? Is there an example of an app that does this?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ParticipantManager.java
```diff
@@ -0,0 +1,111 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+import static org.sagebionetworks.bridge.android.util.retrofit.RxUtils.toBodySingle;
+
+/**
+ * Any authenticated user may use this class's methods. The user does not need to have consented to
+ * the study in order to manage their participant record.
+ * <p>
+ * TODO: offline syncing
+ */
+@AnyThread
```
dephillipsmichael, line 27: I wouldn't suggest using this class from any thread.  That would imply it is ok to edit the AccountDAO from any thread.  Is this your intention?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ConsentManager.java
liujoshua: Good call

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/PlaintextSharedPreferencesDAO.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.Set;
+import java.util.concurrent.ConcurrentMap;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * FIXME: encrypt sensitive information, integrate with fingerprint
+ */
+
+public class PlaintextSharedPreferencesDAO implements AccountDAO, ConsentDAO {
```
dephillipsmichael: I'm not sure you did the class hierarchy of this right.  I'd suggest switching AccountDao and ConsentDao to be there own classes that extend PlaintextSharedPreferencesDao, where that is the class that has the functionality.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
dephillipsmichael: Thanks for the clarification.  In that case, I would just make it always be async and for the sync version, just call the async callback immediately. That way there is only one way.

android-sdk/src/test/java/org/sagebionetworks/bridge/android/manager/auth/InMemoryDAO.java
```diff
@@ -15,7 +18,7 @@
  * Created by jyliu on 1/20/2017.
  */
 @AnyThread
-class InMemoryDAO implements AuthenticationManager.DAO {
+class InMemoryDAO implements AccountDAO {
```
dephillipsmichael: Should this class be deleted?  Seems odd to have two separate classes implementing AccountDao

dephillipsmichael:
Good work overall encapsulating the managers into their corresponding functionality.  I'd like @EricSiegNW or Rian to review as well.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/PlaintextSharedPreferencesDAO.java
liujoshua: Good point

android-sdk/src/test/java/org/sagebionetworks/bridge/android/manager/auth/InMemoryDAO.java
liujoshua: This is used by unit tests to avoid mocking get/set repeatedly or using RobolectricRunner to have LocalStorage. Is there a better approach? It's been several years since I looked

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ParticipantManager.java
liujoshua: What's your opinion on where DAOs should be called?

I intend all the DAOs to be callable from UI and background threads. One benefit of having a DAO (essentially a local cache) is to be able to get a value without making a network call, which would need to go in the background thread



Is this not actually needed?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ConsentManager.java
liujoshua: Yes, I believe so. No, I can't think of one.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ConsentManager.java
liujoshua: Interesting. It appears Bridge  requires name, bday, and sharingScope but not imageData or mimeType. 

It also appears that Bridge checks to make sure you're 18, so I'm thinking I'd want to enforce that in android-sdk before letting it go into local storage... 

This also means that what we thought was a valid consent locally is rejected by Bridge, which I had not considered. This adds some complexity since we can't even actually call Bridge to give consent prior to creating the account... so right now it's possible to do consent workflow (saved offline), sign up, consent gets uploaded in background, and is rejected by Bridge.. user now is unconsented

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ConsentManager.java
liujoshua: Right now mPower doesn't handle the cases associated with the way consents actually work on Bridge, so maybe I could just keep isConsented() for now, or rename to clarify.

Background:

One simplification made in MoleMapper/ResearchStack was assuming one consent, which by convention on Bridge, has a subpopulationGuid = studyId

In reality, a study can have multiple consents, with each consent associated with a subpopulationGuid, each consent can have multiple versions, and be required or not required.

So isConsented() actually needs to look through all required consents and check that the user has consented (to reflect what Bridge believes about whether the user is consented).

isConsented(subpopulation) lets you check for a specific consent

isConsentMostRecent variants takes version into account. I'd imagine this could be useful if the user signed an older version, and therefore has access to Bridge APIs, but governance would really want them to agree to the new version (kinda like new Terms and Conditions for software?)

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/BridgeManagerProvider.java
liujoshua: BridgeManagerProvider is a singleton attached to the application context, so right now I'm initializing BridgeManagerProvider in Application's onCreate(). I'm memoizing because onCreate() specifically suggests lazy loading: https://developer.android.com/reference/android/app/Application.html#onCreate()

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ConsentManager.java
liujoshua: Removed

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ConsentManager.java
liujoshua: I've corrected the annotations and the nullchecks

android-sdk/src/test/java/org/sagebionetworks/bridge/android/manager/auth/InMemoryDAO.java
dephillipsmichael: Ohhh ok, my fault.  The standard for name of objects like this is the "Mock" prefix.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ParticipantManager.java
dephillipsmichael: After reviewing the PlaintextSharedPreferencesDAO it does look like it is thread safe!  I figured you were storing some stuff in member variables, but you aren't.  Looks good to me!  Scratch this comment.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ConsentManager.java
dephillipsmichael: Thanks for taking a second look!

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/BridgeManagerProvider.java
dephillipsmichael: I admire your attention to detail on this!  I do agree that you shouldn't do any heavy lifting in Application.onCreate(). But, on the other hand, that is the correct spot for initializing singletons like BridgeManagerProvider.  And since you are creating the provider object, you should definitely just go ahead and initialize all the managers as well. Creating objects isn't considered anything that will slow the application start up.  That is more like reading data from files.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/PlaintextSharedPreferencesDAO.java
```diff
@@ -0,0 +1,128 @@
...
+
+        return consents.get(subpopulationGuid);
+    }
+
+    public synchronized void put(@NonNull String subpopulationGuid,
+                                 @NonNull ConsentSignature consentSignature) {
+        checkNotNull(subpopulationGuid);
+        checkNotNull(consentSignature);
+
+        consents.put(subpopulationGuid, consentSignature);
```
dephillipsmichael: Isn't "ConcurrentMap<String, ConsentSignature> consents" the same thing as just storing this in SharedPrefs?  You won't have to do any of the synchronization either. My vote would be to push that to shared prefs too

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AccountDAO.java
liujoshua: Made a class

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
liujoshua: Will that work? For example, if during a task or something, the consent needs to be retrieved, does there need to be a version that doesn't do a network call and can be run on the main thread? Or can something like AsyncTask take care of that?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ConsentDAO.java
liujoshua: Removed

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/PlaintextSharedPreferencesDAO.java
liujoshua: Changed to always read/write from shared prefs

liujoshua:
Made some changes that were beyond pure refactoring, so I'll retest onboarding flow Friday. Unfortunately, don't have enough test coverage so manual testing is the only way.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/BridgeManagerProvider.java
```diff
@@ -61,28 +64,36 @@ public static void init(@NonNull Context applicationContext) {
     private BridgeManagerProvider(@NonNull Context applicationContext) {
         this.applicationContext = applicationContext;
         bridgeConfig = new BridgeConfig(this.applicationContext);
-        apiClientProviderSupplier = Suppliers.memoize(this::createApiClientProvider);
-        authenticationManagerSupplier = Suppliers.memoize(this::createAuthenticationManager);
-        participantManagerSupplier = Suppliers.memoize(this::createStudyParticipantManager);
-        consentManagerSupplier = Suppliers.memoize(this::createConsentManager);
-        plaintextSharedPrefsDAOSupplier = Suppliers.memoize(this::createSharedPreferencesDAO);
 
+        apiClientProvider =  new ApiClientProvider(
```
dephillipsmichael: Thank you :)

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ConsentManager.java
```diff
@@ -157,38 +195,19 @@ public Completable giveConsent(@NonNull String subpopulationGuid, @NonNull Strin
...
+        return consentSignature;
     }
 
-    /**
-     * Sends a copy of the participant's signed consent to their email address.
-     *
-     * @param subpopulationGuid guid for the subpopulation of the consent
-     * @return completable
-     */
-    @NonNull
```
dephillipsmichael: Did you delete the "saveLocally" method from DataProvider?  I think this would be a good time to do that refactor.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/dao/AccountDAO.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.sagebionetworks.bridge.rest.model.SignIn;
+import org.sagebionetworks.bridge.rest.model.StudyParticipant;
+import org.sagebionetworks.bridge.rest.model.UserSessionInfo;
+
+/**
+ * Created by jyliu on 2/8/2017.
+ */
+@AnyThread
+public class AccountDAO extends SharedPreferencesJsonDAO {
```
dephillipsmichael, line 15: Nice this is super clean

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/dao/ConsentDAO.java
```diff
@@ -0,0 +1,95 @@
...
+
+        logger.debug("put called for subpopulations " + subpopulationGuid
+                + ", with: " + consentSignature);
+
+        Map<String, ConsentSignature> consents = load();
+        consents.put(subpopulationGuid, consentSignature);
+        persist(consents);
+    }
+
+    public synchronized void remove(@NonNull String subpopulationGuid) {
```
dephillipsmichael: I don't think "synchronized" is need anymore right?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/dao/ConsentDAO.java
```diff
@@ -0,0 +1,95 @@
...
+        Map<String, ConsentSignature> consents = load();
+        ConsentSignature consentSignature = consents.get(subpopulationGuid);
+
+        logger.debug("get called for subpopulation " + subpopulationGuid
+                + ", found: " + consentSignature);
+
+        return consentSignature;
+    }
+
+    public synchronized void put(@NonNull String subpopulationGuid,
```
dephillipsmichael: You can certainly store a map, but it's easier just to store the data in SharedPrefs with "subpopulationGuid" as the key, and the ConsentSignature as the value.  And then when you wanted the "map" at that point, you would just call sharedPrefs.getAll() which returns a Map<String, ?>

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -47,27 +47,23 @@
...
 
-    protected final Gson gson = RestUtils.GSON;
-
     // set in initialize
-    protected UserLocalStorage userLocalStorage;
-    protected ConsentLocalStorage consentLocalStorage;
     protected TaskHelper taskHelper;
     protected UploadHandler uploadHandler;
 
+    private final StorageAccessWrapper storageAccessWrapper;
```
dephillipsmichael: What's with StorageAccessWrapper?  The header says it is to help unit tests, but I think it just creates another layer of abstract confusion, especially since StorageAccess is a singleton.  The point of a singleton is there is only one access point for the class, but then we pass it around in a wrapper?  Why not just pass around StorageAccess at that point?

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -77,25 +73,29 @@
...
         this.authenticationManager = bridgeManagerProvider.getAuthenticationManager();
         this.consentManager = bridgeManagerProvider.getConsentManager();
-
     }
 
     public BridgeDataProvider(BridgeManagerProvider bridgeManagerProvider) {
+        this.researchStackDAO = new ResearchStackDAO(bridgeManagerProvider.getApplicationContext());
         this.bridgeManagerProvider = bridgeManagerProvider;
-
+        this.storageAccessWrapper = new StorageAccessWrapper();
```
dephillipsmichael: Hmmmm I can't imagine this is a good idea.

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -77,25 +73,29 @@
...
-
     }
 
 
     @Override
     public Observable<DataResponse> initialize(Context context) {
         logger.debug("Called initialize");
 
-        // TODO: data provider
+        // temporary hard coded, this will no longer be needed with StudyUploadEncrypter
```
dephillipsmichael: add the TODO back in

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -232,13 +235,28 @@ protected ConsentSignature createConsentSignatureBody(TaskResult consentResult)
 
         // Save Consent Information
         // User is not signed in yet, so we need to save consent info to disk for later upload
-        return new ConsentSignature().name(fullName).birthdate(new LocalDate(birthdateInMillis)).imageData(base64Image).imageMimeType("image/png").scope(SharingScope.valueOf(sharingScope));
+        return new ConsentSignature()
+                .name(fullName)
+                .birthdate(new LocalDate(birthdateInMillis))
+                .imageData(base64Image)
+                .imageMimeType("image/png")
```
dephillipsmichael, line 242: hard-coded image/png?

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -430,24 +423,18 @@ public Completable forgotPassword(@NonNull String email) {
...
-        logger.debug("Called getUser");
-        userLocalStorage.saveUser(user);
+        researchStackDAO.setUser(user);
     }
 
     @Override
     public String getUserEmail(Context context) {
-        User user = userLocalStorage.loadUser();
-        return user == null ? null : user.getEmail();
+        return authenticationManager.getEmail();
```
dephillipsmichael: Seems odd that email comes from the auth manager, shouldn't it be stored in the user object?

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
dephillipsmichael: What's the difference between ResearchStackDAO which contains User, and the AccountDAO?  They seem like they are duplicated in some ways.

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/ResearchStackDAO.java
```diff
@@ -0,0 +1,28 @@
...
+import android.content.Context;
+
+import org.researchstack.backbone.model.User;
+import org.sagebionetworks.bridge.android.manager.dao.SharedPreferencesJsonDAO;
+
+/**
+ * Created by jyliu on 2/9/2017.
+ */
+
+public class ResearchStackDAO extends SharedPreferencesJsonDAO {
```
dephillipsmichael, line 12: What's the difference between ResearchStackDAO which contains User, and the AccountDAO? They seem like they are duplicated in some ways.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/dao/ConsentDAO.java
liujoshua: That sounds way too straightforward to me ;)

dephillipsmichael: hahaha spoken like a true programmer

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/dao/AccountDAO.java
liujoshua: Thanks

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
liujoshua: This was created prior to having more ability to refactor ResearchStack to isolate dependencies that force a call to Android APIs, which all throw RuntimeException("Stub!") without Robolectric (which is much slower than normal unit tests)

StorageAccess performs Android calls on class load, so if you call  mock/createMock(StorageAccess.class), it will barf on this line (IIRC):
    private Handler handler = new Handler(Looper.getMainLooper());

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
liujoshua: This is what ConsentSignatureStepLayout returns and there's no other use case that I know of, so I think it can be deferred. I'll create a JIRA

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
liujoshua: I don't know.. should getUserEmail return the signed in user? or the user in the User object?

I don't understand the use case for storing User but added it to not regress what's currently in RS. AFAIK this didn't exist in MoleMapper or ResearchStack up until recently, User object was not used at all by RS before

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/ResearchStackDAO.java
liujoshua: AccountDAO/ConsentDAO works with Bridge objects (android-sdk, no dependency on RS) and ResearchStackDAO works with objects in RS.

Also, I don't understand the user case for User, but it seems distinct from the Bridge concept of account

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ConsentManager.java
liujoshua: I'd love to, but may have to wait until I get back unless someone else does it. The code change should be trivial at this point, but I don't think I'll have time to test it sufficiently

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/dao/ConsentDAO.java
liujoshua: removed

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/ResearchStackDAO.java
liujoshua: More on the User issue. So I want to do what's correct, but we should clarify the meaning and use cases for Session & SignIn & StudyParticipant (Bridge concepts) and User.

My thinking when I was working on this is that User mostly duplicates what's already in AuthManager and StudyParticipant, adding in birth date. And then, it seems the use of birth date is while signing consent, if you've already populated a birth date by signing a consent?

It looks like you most recently touched the User stuff, could you clarify what it's for?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/BridgeManagerProvider.java
```diff
@@ -1,52 +1,157 @@
...
+                bridgeConfig.getAcceptLanguage());
+
+        accountDAO = new AccountDAO(applicationContext);
+        consentDAO = new ConsentDAO(applicationContext);
+
+        authenticationManager = new AuthenticationManager(bridgeConfig, apiClientProvider, accountDAO);
+        participantManager = new ParticipantManager(authenticationManager, accountDAO);
+        consentManager = new ConsentManager(authenticationManager, consentDAO);
+
+        try {
```
liujoshua, line 77: We went away from lazy loading, so I'll point out this does load from assets here (perhaps that's non-trivial work?) My intention is that this will crash the app early if there will be a problem encrypting data archives so I'm wrapping in an unchecked exception and rethrowing. StudyUploadEncryptor gives a meaningful error message, hence the lame one here.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ProxiedForConsentedUsersApi.java
```diff
@@ -31,6 +31,7 @@
 /**
  * Proxies ForConsentedUserApi calls so it uses retrieves authenticated client from AuthenticationManager
  * to make the call.
+ * TODO: push refreshing of credentials into rest-client's authentication handler
  */
 class ProxiedForConsentedUsersApi implements ForConsentedUsersApi {
```
rianhouston, line 36: I don't really understand the need for this class.  I thought this was handled already by the rest-client.  It seems very odd to have to reimplement all the methods this way.  Can you provide a little more explanation?

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -18,569 +16,503 @@
...
+    }
+
+    @NonNull
+    public Completable withdrawConsent(@NonNull String subpopulationGuid, @Nullable String reason) {
+        return consentManager.withdrawConsent(subpopulationGuid, reason);
+    }
+
+
+    @NonNull
+    public Completable withdrawAllConsents(@Nullable String reason) {
```
rianhouston, line 125: Are these methods intended for external use?  It seems as though it is just an internal helper to manage the return type differences between DataProvider and the managers?  Should it be public?

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/ResearchStackDAO.java
rianhouston: Yeah, this is confusing to me also.  We have model classes defined by the rest-client and model classes defined in RS.  It seems like we need a translation between the two?

rianhouston:
Overall, this looks great.  It makes a lot more sense to me now than the last PR I reviewed.  There are still a few things unclear to me but I would be ok merging this and building from here.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ProxiedForConsentedUsersApi.java
liujoshua: Yes, this is odd, At the very least, it belongs in rest-client, as you mentioned. I've added the TODO and created a JIRA to try and point out it's a known issue. 

rest-client handles reauthentication, but ForConsentedUsersApi it offers is bound to an email-password pair. This proxy binds ForConsentedUsersApi to the email-password pair held in AuthenticationManager. With the unproxied client, any class using the API would need to discard instances of ForConsentedUsersApi (or not hold onto to begin with) if the user signs in or out, orpassword changes. An instance of ProxiedForConsentedUsersApi will by synced with email/password called on DataProvider.signIn/signOut.

Certainly, this type of functionality belongs in rest-client. But I think perhaps it's also doing a bit more than it needs to. I could see wanting to be able to reuse the instance for same account (email) across password changes, but needing to reset state or create new objects in the event of a logout/login. 

IMO it's ugly and perhaps does more than it needs, but it does work to keep the Api/AuthMan in sync. I suggest we defer this change for now and figure out the cleanest/best solution when we pick up the JIRA.

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/ResearchStackDAO.java
liujoshua: I agree this needs further refinement. I do think translation is appropriate like you mention. The android-sdk should work with Bridge models only, and Bridge models shouldn't and don't make it outside of DataProvider. 

However, a simple adapter didn't work in this case because the Bridge models don't contain enough information to create the User object in RS, hence an additional storage in researchstack-sdk. It seemed all l I needed in addition to Bridge models was birthday, but the use case for User was unclear to me. So I put in this short term fix of an additional storage layer for things needed by RS but unknown to Bridge to not break functionality

liujoshua:
Thanks :) I hope things continue getting clearer and cleaner.

I came into this with design plans for how android-sdk and researchstack-sdk would work but am definitely learning as I go. For example, I have a pretty good grasp on account management and the data upload side of things, but I'll basically be starting from scratch when it comes the activities and how they relate to RS when I start on the next chunks of features.

Also, there are things I'd like to refactor between the RS, android-sdk, and researchstack-sdk packages which will further clarify things. But I think I should wait until the required areas are built and there's some more test coverage before attempting that. Me breaking things for you guys is inevitable, but I'd prefer to minimize it :)

dephillipsmichael (approved):
Some loose ends to tie up still, but overall this is looking good.

PR #30, opened 2017-02-13
ActivityManager

rianhouston:
This is a WIP but is the first step to retrieve activities/tasks from the webservice rather than read from a file.

### DIFF
--- a/researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
+++ b/researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
@@ -26,6 +26,8 @@ import org.sagebionetworks.bridge.android.manager.BridgeManagerProvider;
 import org.sagebionetworks.bridge.android.manager.ConsentManager;
 import org.sagebionetworks.bridge.researchstack.wrapper.StorageAccessWrapper;
 import org.sagebionetworks.bridge.rest.model.ConsentSignature;
+import org.sagebionetworks.bridge.rest.model.ScheduledActivity;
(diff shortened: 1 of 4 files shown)


liujoshua (approved):

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -484,11 +489,16 @@ public void setUserSharingScope(Context context, String scope) {
 
     @Override
     public SchedulesAndTasksModel loadTasksAndSchedules(Context context) {
-        // TODO: integrate with bridge
-        // forConsentedUsersApi.getSchedules();
-        // forConsentedUsersApi.getScheduledActivities();
+        logger.info("loadTasksAndSchedules()");
+
+        // TODO: figure out the correct arguments to pass here
+        ScheduledActivityList scheduledActivityList = bridgeManagerProvider.getActivityManager().getActivities("-07:00", 1, 0)
```
dephillipsmichael, line 495: Looks like the argument in iOS is a string version of today's date offset formatted as ISO8601 which is like '+00:00'. 

However, iOS doesn't have that as a parameter to pass in at all so I would just compute it before making the call.

dephillipsmichael:
Looks good to me too

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
rianhouston: Totally agree.  I have already made that change locally and will include in the next update.

PR #32, opened 2017-03-05
feature/ma-75

dephillipsmichael:
This PR is dependent on https://github.com/ResearchStack/ResearchStack/pull/357

Updates to Bridge to support a wider variety of Results for upload.  Some of the changes are simply switching the indentation from 2 to 4 spaces for a tab; however, a full review of this code should be helpful.


researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/TaskHelper.java
```diff
@@ -2,181 +2,329 @@
...
+            // loading the task json here is bad, but the taskID is in the schedule
+            // json but the readable id is in the task json
+            TaskModel taskModel = loadTaskModel(context, task);
+            TaskResult result = db.loadLatestTaskResult(taskModel.identifier);
+
+            // cache cron string for later lookup
+            loadedTaskCrons.put(taskModel.identifier, schedule.scheduleString);
+
+            if (result == null) {
+                schedules.add(schedule);
```
liujoshua, line 104: This file has changes even though github is collapsing it

researchstack-sdk/build.gradle
```diff
@@ -55,5 +55,7 @@ dependencies {
   testCompile 'junit:junit:4.12'
   testCompile 'org.robolectric:robolectric:3.0'
   testCompile 'org.mockito:mockito-core:1.10.19'
+  testCompile 'org.powermock:powermock-api-mockito:1.6.6'
+  testCompile 'org.powermock:powermock-module-junit4:1.6.6'
```
dephillipsmichael, line 59: I just added these to backbone, I think that they should be available in researchstack-sdk without specifically adding them as a dependency.

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/TaskHelper.java
dephillipsmichael: Yeah I saw that, I wonder why it does that, it's very confusing.

dephillipsmichael:
Looks good to me.  You did a few things I was wanting to do with the model for Archives.  Nice work.

researchstack-sdk/build.gradle
liujoshua: Hmm.. didn't work for me after a clean/build. Are test dependencies supposed to be transitive in Gradle?

PR #37, opened 2017-03-16
Use real schemaId; Add debug logging

liujoshua:
Depends on this PR: https://github.com/ResearchStack/ResearchStack/pull/362/files

### DIFF
--- a/android-sdk/src/main/java/org/sagebionetworks/bridge/android/util/retrofit/RxUtils.java
+++ b/android-sdk/src/main/java/org/sagebionetworks/bridge/android/util/retrofit/RxUtils.java
@@ -6,6 +6,7 @@ import android.support.annotation.NonNull;
 import retrofit2.Call;
 import rx.Observable;
 import rx.Single;
+import rx.schedulers.Schedulers;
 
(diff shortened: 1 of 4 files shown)


android-sdk/src/main/java/org/sagebionetworks/bridge/android/util/retrofit/RxUtils.java
```diff
@@ -14,11 +15,20 @@
...
+     * @return Single created from call, which emits T
+     * @typeparam T type returned by Single and Call
+     */
     @NonNull
     public static <T> Single<T> toBodySingle(@NonNull Call<T> call) {
         checkNotNull(call);
 
-        return Observable.create(new BodyOnSubscribe<>(new CallOnSubscribe<>(call))).toSingle();
+        return Observable.create(new BodyOnSubscribe<>(new CallOnSubscribe<>(call)))
+                .subscribeOn(Schedulers.io()).toSingle();
```
liujoshua, line 31: This makes all Retrofit Call<T> happen on the io(). That way we won't have to think about thread every time we do a network call

dephillipsmichael (approved):
Looks good! approved.

PR #39, opened 2017-03-20
Feature/upload

liujoshua:
Queue pending uploads by saving them locally for future processing


dephillipsmichael:
Unit tests need updated to reflect new queue/upload flow.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
```diff
@@ -26,51 +28,145 @@
...
 
     private final ForConsentedUsersApi api;
     private final StudyUploadEncryptor encryptor;
-    private final OkHttpClient s3OkhttpClient;
+    private final UploadDAO uploadDAO;
+    private final OkHttpClient okHttpClient = new OkHttpClient.Builder()
+            .connectTimeout(30, TimeUnit.SECONDS)
+            .readTimeout(30, TimeUnit.SECONDS)
+            .writeTimeout(30, TimeUnit.SECONDS)
+            .retryOnConnectionFailure(false).build();
```
dephillipsmichael, line 78: It seems like this would be a nice way to get free retry logic.  Did you consider enabling it?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
```diff
@@ -26,51 +28,145 @@
...
+                });
     }
 
-    @NonNull
-    Single<UploadValidationStatus> uploadToS3(UploadFile uploadFile, UploadSession session) {
+    // determines whether an upload failure is permanent, i.e. no point in retrying
+    boolean isPermanentFailure(UploadFile uploadFile, Throwable throwable) {
+        if (throwable instanceof IllegalStateException) {
+            // some pre-condition failed
+            return true;
```
dephillipsmichael: Can you elaborate more in comments about specifically what would cause an IllegalStateException here?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
```diff
@@ -26,51 +28,145 @@
...
+            return Single.just(null);
+        }
 
-        return Single.zip(uploadFileSingle, uploadSessionSingle, this::uploadToS3).flatMap(i -> i);
+        return getUploadSession(uploadFile)
+                .flatMap(uploadSession -> getUploadValidationStatus(uploadSession.getId()))
+                .flatMap((uploadValidationStatus) -> {
+                    if (uploadValidationStatus == null) {
+                        LOG.info("Could not find uploadValidationStatus for filename: " + filename);
+                        return Single.just(null);
```
dephillipsmichael: Should this be an error message of some sort?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
```diff
@@ -26,51 +28,145 @@
...
+                        uploadDAO.removeUploadAndSession(filename);
+                        return Single.just(null);
+                    } else if (UploadStatus.VALIDATION_FAILED == uploadValidationStatus.getStatus()) {
+                        LOG.info("Cleaning up validation-failed upload with filename: " + filename
+                                + ", id: " + uploadValidationStatus.getId());
+                        uploadDAO.removeUploadAndSession(filename);
+                        return Single.just(null);
+                    }
+
+                    return getUploadSession(uploadFile)
```
dephillipsmichael: This return is crazy complex.  Can you add comments on why you would do the entire upload flow again here?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
```diff
@@ -26,51 +28,145 @@
...
+                            .flatMapCompletable(uploadSession -> uploadToS3(uploadFile, uploadSession))
+                            .andThen(getUploadValidationStatus((uploadValidationStatus.getId())));
+                }).doOnSuccess(uploadValidationStatus -> {
+                    LOG.debug("uploadToS3 succeded, uploadValidationStatus: " + uploadValidationStatus);
```
dephillipsmichael: Wouldn't this make more sense to Log in the block above?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
```diff
@@ -104,24 +203,28 @@ public UploadManager(AuthenticationManager authenticationManager, StudyUploadEnc
...
-                    // TODO: insert upload session to DB, link to file metadata
-                    LOG.info("Received upload session with id: " + uploadSession.getId());
+                    LOG.info("Received uploadToS3 session with id: " + uploadSession.getId());
+                    uploadDAO.putUploadSession(uploadFile.filename, uploadSession);
                 });
     }
 
-    UploadFile persist(String filename, Archive archive) throws NoSuchAlgorithmException, IOException {
+    @WorkerThread
+    UploadFile persist(String filename, Archive archive) throws IOException,
```
dephillipsmichael: Can you put method headers and descriptions on all the methods in this class?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/BridgeManagerProvider.java
```diff
@@ -113,6 +117,8 @@ private BridgeManagerProvider(@NonNull Context applicationContext) {
     @NonNull
     private final AccountDAO accountDAO;
     @NonNull
+    private final UploadDAO uploadDAO;
```
dephillipsmichael: How/when does the uploadDAO get cleared out after sign out?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/util/retrofit/RxUtils.java
```diff
@@ -18,7 +19,8 @@
     public static <T> Single<T> toBodySingle(@NonNull Call<T> call) {
```
dephillipsmichael, line 27: Can you add a method header describing this?

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/TaskHelper.java
```diff
@@ -189,19 +188,25 @@ void uploadTaskResult(TaskResult taskResult, Archive.WithBridgeConfig withBridge
...
 
-        // At this point, the upload request has been processed and saved,
-        // so it is safe to delete the temporary data logger files
-        for (Result result : results) {
-            if (result instanceof FileResult) {
-                FileResult fileResult = (FileResult) result;
-                DataLoggerManager.getInstance().deleteFileStatus(fileResult.getFile());
-            }
-        }
+                    bridgeManagerProvider.getUploadManager().uploadToS3().toCompletable().await();
```
dephillipsmichael: This would imply that this method uploadTaskResult, needs to be off the main thread correct?  Can you add a method header/description/annotation to reflect this?

Also, if this is on a different thread, why not just do ".await()" for the queue upload as well?

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/TaskHelper.java
```diff
@@ -189,19 +188,25 @@ void uploadTaskResult(TaskResult taskResult, Archive.WithBridgeConfig withBridge
...
+                .queueUpload(archiveFilename, bridgeArchiveBuilder.build())
+                .doOnSuccess(uploadFile -> {
+                    logger.debug("Calling uploadToS3() to upload pending uploads");
+
+                    // At this point, the upload request has been processed and saved,
+                    // so it is safe to delete the temporary data logger files
+                    for (Result result : results) {
+                        if (result instanceof FileResult) {
+                            FileResult fileResult = (FileResult) result;
+                            DataLoggerManager.getInstance().deleteFileStatus(fileResult.getFile());
```
dephillipsmichael: I'm not entirely sure if deleteFileStatus is thread-safe.  I didn't intend on it to be called off the  main thread.

dephillipsmichael:
IMO, it's hard to follow the logic of the UploadManager, which is making it difficult to truly review.  The iOS version, SBBUploadManager, is much easier to understand since it has detailed documentation, and specifically deals with all success / http error code responses.  I'd suggest adding detailed comments to your code before I can approve, using iOS as a reference.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/util/retrofit/RxUtils.java
liujoshua: Added

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/BridgeManagerProvider.java
liujoshua: It currently doesn't. I created a separate story/sub-task in JIRA for this

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
liujoshua: Good idea. I have a branch that moves the call to the network part of the upload into a SyncProvider, which I believe handles connection failures. I'll test it with this flag

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/BridgeManagerProvider.java
dephillipsmichael: I think that should be a part of this story.  Seems like a really easy task, no?  On successful sign out, call clear.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
```diff
@@ -26,76 +30,257 @@
...
-                .fromCallable(() -> persist(filename, archive)).cache();
+    public Single<UploadFile> queueUpload(String filename, Archive archive) {
+        return Single.fromCallable(() -> persist(filename, archive))
+                .subscribeOn(Schedulers.io());
+    }
 
-        Single<UploadSession> uploadSessionSingle = uploadFileSingle
-                .flatMap(this::requestUploadSession);
+    /**
+     * This triggers an attempt to upload all the previously queued upload files
```
dephillipsmichael: How many threads does the Schedulers.io() have in it's thread pool?  This might be important to note in the comments.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
```diff
@@ -26,76 +30,257 @@
...
+     * You must call queueUpload before this method will have anything to operate on
+     *
+     * @return Observable with information on if the upload was successful or not
```
dephillipsmichael: "Observable with information on if the upload was successful or not"... which upload?  all of them?  Seems like this might make sense as a Void return type if you can't return the status of all of them as a list.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
```diff
@@ -26,76 +30,257 @@
...
+
+        return filenameObservable
+                .map(filename -> uploadDAO.getUploadFile(filename))
+                .flatMap(uploadFile -> processUploadFile(uploadFile).toObservable())
+                .toCompletable();
+    }
 
-        return Single.zip(uploadFileSingle, uploadSessionSingle, this::uploadToS3).flatMap(i -> i);
+    /**
+     * Performs next upload step for a file.
```
dephillipsmichael: Honestly, this doesn't really help me understand what is going on in the method.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
```diff
@@ -26,76 +30,257 @@
...
+import rx.Observable;
 import rx.Single;
+import rx.schedulers.Schedulers;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
  * Created by jyliu on 1/30/2017.
  */
```
dephillipsmichael: I see a lot of references to "the next step in the upload process" so i think it would be good to give an overview of the full upload process here.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
```diff
@@ -26,76 +30,257 @@
...
+                sessionSingle,
+                statusSingle,
+                this::processUploadForValidationStatus
+        ).flatMapCompletable(i -> i);
     }
 
+    /**
+     * Performs the next upload step, according to the UploadValidationStatus. The normal
+     * transition of UploadValidationStatus is from REQUESTED to VALIDATION_IN_PROGRESS to
+     * REQUESTED.
```
dephillipsmichael: do you mean to COMPLETE at the end?

dephillipsmichael: sorry i mean SUCCESS

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
```diff
@@ -26,76 +30,257 @@
...
+                return uploadToS3(uploadFile, uploadSession);
+            case SUCCEEDED:
+            case DUPLICATE:
+                return dequeueUpload(uploadFile.filename);
+            case VALIDATION_IN_PROGRESS:
+                LOG.debug("Validation in progress for filename: " + uploadFile.filename +
+                        ", uploadId" + uploadSession.getId());
+                break;
+            case VALIDATION_FAILED:
+                LOG.debug("Validation failed for filename: " + uploadFile.filename +
```
dephillipsmichael, line 242: Should this dequeueUpload as well?  Or will this fail silently and it will be attempted again?

dephillipsmichael: Never mind, the comment at the end answer my questions, nice!

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
```diff
@@ -26,76 +30,257 @@
...
+    Completable dequeueUpload(@NonNull String filename) {
+        checkNotNull(filename, "filename required");
+
+        if (getFile(filename).delete()) {
+            LOG.warn("Successfully deleted upload file: " + filename + ", "
+                    + "removing upload from queue");
+            return Completable.fromAction(() -> uploadDAO.removeUploadAndSession(filename));
+        } else {
+            LOG.warn("Failed to delete upload file: " + filename);
+            return Completable.error(new IOException("Failed to delete upload file: " + filename));
```
dephillipsmichael: I usually like to make delete file errors log but not actually throw an error like the completable error, because usually they mean the file didn't exists to delete and that is like a success

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
```diff
@@ -26,76 +30,257 @@
...
+     */
+    @NonNull
+    Single<UploadValidationStatus> getUploadValidationStatus(@NonNull String uploadId) {
+        checkNotNull(uploadId, "uploadId required");
+
+        return RxUtils.toBodySingle(api.getUploadStatus(uploadId));
+    }
+
+    @NonNull
+    Completable uploadToS3(UploadFile uploadFile, UploadSession session) {
```
dephillipsmichael: method header?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
```diff
@@ -26,76 +30,257 @@
...
         File file = getFile(uploadFile.filename);
+        checkArgument(file.exists(), "Non-existent file: " + file.getAbsolutePath());
+
+        Single<UploadSession> sessionSingle = Single.just(session)
```
dephillipsmichael, line 300: What is going on here?  You take an UploadSession, and wrap it in a Single and then operate on that same variable?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
```diff
@@ -26,76 +30,257 @@
...
+                getS3Service(freshSession)
+                        .uploadToS3(
+                                freshSession.getUrl(),
+                                requestBody,
+                                uploadFile.md5Hash,
+                                uploadFile.contentType)))
+                .doOnSuccess(aVoid -> {
+                    LOG.info("S3 upload succeeded for id: " + session.getId());
+
+                    // call upload complete on a computation thread
```
dephillipsmichael, line 325: Seems odd you chose this for a computation thread when all the other network calls happen on IO.  What was your reasoning?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/dao/UploadDAO.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Set;
+
+/**
+ * Created by jyliu on 3/10/2017.
+ */
+
+public class UploadDAO extends SharedPreferencesJsonDAO {
+    private static final Logger logger = LoggerFactory.getLogger(UploadDAO.class);
+
+    private static final String PREFERENCES_FILE  = "uploads";
```
dephillipsmichael, line 21: The more I think about it, we should move all the DAOs form shared preferences to StorageAccess.  That way we can just delete all the files in StorageAccess' FileAccess when the user signs out and it will make life a lot easier.

android-sdk/src/test/java/org/sagebionetworks/bridge/android/manager/UploadManagerTest.java
```diff
@@ -0,0 +1,204 @@
...
+
+    private UploadManager spyUploadManager;
+
+    private UploadManager.UploadFile uploadFile;
+    private UploadSession uploadSession;
+    private UploadValidationStatus uploadValidationStatus;
+
+
+    @Before
+    public void setupTest() {
```
dephillipsmichael, line 69: Nice test coverage on this file

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/TaskHelper.java
```diff
@@ -195,16 +198,25 @@ void uploadTaskResult(TaskResult taskResult, Archive.WithBridgeConfig withBridge
...
-        for (Result result : results) {
-            if (result instanceof FileResult) {
-                FileResult fileResult = (FileResult) result;
-                DataLoggerManager.getInstance().deleteFileStatus(fileResult.getFile());
-            }
-        }
+                .queueUpload(archiveFilename, bridgeArchiveBuilder.build())
+                .doOnSuccess(uploadFile -> {
+                    logger.debug("Attempting upload in io() thread");
+                    bridgeManagerProvider.getUploadManager()
```
dephillipsmichael: If upload needs to happen on the IO thread, I think the upload manager should put it on the I/O thread, since it shouldn't be the responsibility of the caller to know to call process upload file on a specific thread.

dephillipsmichael:
The code is a lot easier to follow now, thank you for adding all the documentation and re-organizing so that is more readable.  I still have some issues from my comments, but I was able to understand everything fully.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
liujoshua: These schedulers are optimized/managed for us by the platform/library.

The comment on Schedulers.io() indicate it's backed by an Executor thread-pool that will grow as needed.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
liujoshua: Good point

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/BridgeManagerProvider.java
liujoshua: Yea that makes sense, I'll clear all for now, but that is not the full story. When the user signs out, or a different user signs in, files waiting for upload and already have a UploadSession (pre-signed URL) should still get uploaded.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
liujoshua: I did change it to an Observable<Void>, but I changed it back to Completable because Observable<Void> didn't provide an advantage (can't do anything with it). I did some googling on Observable<Void>, Completables, and Observable<Completable> and found a more readable and useful way to run a collection of Completables: Completable.mergeDelayError

(see https://github.com/ReactiveX/RxJava/issues/3807)

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
liujoshua: Yea that sounds better

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
liujoshua: Fixed, you're right I didn't mean to do that. TUrns out, my attempt to run it on computation thread wouldn't have worked, because RxUtils.toBodySingle already called subscribeOn(Schedulers.io())

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
liujoshua: If the session is not expired or nearing expiry, yes, it is the identity function, we just reuse the Single<Session> created from the value we were passed as a param. Otherwise, the Single<Session> is the result of a request to Bridge for a new Session.

The Single<Session> (containing either the Session in param, or new one for Bridge) is then used for the S3Upload.

Do you think it would be easier to test/understand if i split it up?

So instead of uploadToS3(UplaodFile,UploadSession) that refreshes session and uploads, two separate methods, something like

refreshStaleSession(UploadSession session)

uploadToS3(UploadFile file, String url)

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/dao/UploadDAO.java
liujoshua: Storage Access is in RS, android-sdk has no RS dependency. You are on the right track that researchstack-sdk should implement StorageAccess or otherwise hook in the DAOs.

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/TaskHelper.java
liujoshua: I agree, I believe this line isn't needed. UploadManager is already handling the threading.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
```diff
@@ -193,11 +205,25 @@ public Completable resendEmailVerification(@NonNull final String email) {
...
+        }
+
         return RxUtils.toBodySingle(authenticationApi.signOut())
                 .doOnSuccess(message -> {
                     accountDAO.setSignIn(null);
                     accountDAO.setUserSessionInfo(null);
                     accountDAO.setStudyParticipant(null);
+
+                    for (AuthenticationEventListener listener : listeners) {
+                        listener.onSignedOut(email);
```
dephillipsmichael, line 225: It's usually not the best practice to call listeners on a background thread.  Can we post these to the main thread?

PR #41, opened 2017-03-22
Feature/activity archive creation

liujoshua:
These are some building blocks for identifying the schema for a TaskResult in order to produce a properly formatted archive. The intent is to have allow Schema and Survey formats to be configured by the app, as well as send down from Bridge.

### DIFF
--- a/android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ActivityManager.java
+++ b/android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ActivityManager.java
@@ -2,7 +2,7 @@ package org.sagebionetworks.bridge.android.manager;
 
 import android.support.annotation.NonNull;
 
-import org.sagebionetworks.bridge.android.manager.activity.ActivityCache;
+import org.sagebionetworks.bridge.android.manager.activity.ActivitySchemaCache;
 import org.sagebionetworks.bridge.rest.api.ForConsentedUsersApi;
 import org.sagebionetworks.bridge.rest.model.ScheduledActivity;
 import org.sagebionetworks.bridge.rest.model.ScheduledActivityList;
@@ -31,9 +31,9 @@ public class ActivityManager {
     private final ForConsentedUsersApi api;
 
     @NonNull
-    private final ActivityCache activityCache;
+    private final ActivitySchemaCache activityCache;
 
-    public ActivityManager(@NonNull AuthenticationManager authenticationManager, ActivityCache activityCache) {
+    public ActivityManager(@NonNull AuthenticationManager authenticationManager, ActivitySchemaCache activityCache) {
         checkNotNull(authenticationManager);
 
         api = authenticationManager.getApi();
@@ -52,7 +52,7 @@ public class ActivityManager {
                 scheduleActivityList -> {
                     LOG.debug("Got scheduled activity list");
                     for (ScheduledActivity scheduleActivity : scheduleActivityList.getItems()) {
-                        activityCache.cacheActivity(scheduleActivity.getActivity());
+                        activityCache.cacheActivitySchema(scheduleActivity.getActivity());
                     }
                 });
     }
--- a/android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ActivityManager.java
+++ b/android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ActivityManager.java
@@ -2,6 +2,7 @@ package org.sagebionetworks.bridge.android.manager;
 
 import android.support.annotation.NonNull;
 
+import org.sagebionetworks.bridge.android.manager.activity.ActivityCache;
 import org.sagebionetworks.bridge.rest.api.ForConsentedUsersApi;
 import org.sagebionetworks.bridge.rest.model.ScheduledActivity;
 import org.sagebionetworks.bridge.rest.model.ScheduledActivityList;
@@ -29,24 +30,31 @@ public class ActivityManager {
     @NonNull
     private final ForConsentedUsersApi api;
 
+    @NonNull
+    private final ActivityCache activityCache;
 
-    public ActivityManager(@NonNull AuthenticationManager authenticationManager) {
+    public ActivityManager(@NonNull AuthenticationManager authenticationManager, ActivityCache activityCache) {
         checkNotNull(authenticationManager);
 
-        this.api = authenticationManager.getApi();
+        api = authenticationManager.getApi();
+        this.activityCache = activityCache;
     }
 
 
+    @NonNull
+    public Single<ScheduledActivityList> getActivities(int daysAhead, int minimumPerSchedule) {
+        return getActivities(getTimezoneOffset(), daysAhead, minimumPerSchedule);
+    }
+
+    @NonNull
     public Single<ScheduledActivityList> getActivities(String offset, int daysAhead, int minimumPerSchedule) {
         return toBodySingle(api.getScheduledActivities(offset, daysAhead, minimumPerSchedule)).doOnSuccess(
                 scheduleActivityList -> {
                     LOG.debug("Got scheduled activity list");
+                    for (ScheduledActivity scheduleActivity : scheduleActivityList.getItems()) {
+                        activityCache.cacheActivity(scheduleActivity.getActivity());
+                    }
                 });
-
-    }
-
-    public Single<ScheduledActivityList> getActivities(int daysAhead, int minimumPerSchedule) {
-        return getActivities(getTimezoneOffset(), daysAhead, minimumPerSchedule);
     }
 
     public Completable updateActivities(@NonNull List<ScheduledActivity> scheduledActivities) {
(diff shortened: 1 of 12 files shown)


android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ActivityManager.java
```diff
@@ -29,24 +30,31 @@
...
         checkNotNull(authenticationManager);
 
-        this.api = authenticationManager.getApi();
+        api = authenticationManager.getApi();
+        this.activityCache = activityCache;
     }
 
 
+    @NonNull
+    public Single<ScheduledActivityList> getActivities(int daysAhead, int minimumPerSchedule) {
```
dephillipsmichael, line 45: Can you add method headers to both the "getActivities" methods

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/activity/ActivitySchemaCache.java
```diff
@@ -0,0 +1,156 @@
...
+    /**
+     * Caches the activity's schema information according to its type.
+     *
+     * @param activity activity to cache
+     */
+    public void cacheActivitySchema(Activity activity) {
+        switch (activity.getActivityType()) {
+            case TASK:
+                TaskReference taskReference = activity.getTask();
+                if (taskReference != null) {
```
dephillipsmichael, line 84: You should log an error of taskReference is null here.  Along with the other checks in this switch.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/activity/ActivitySchemaCache.java
```diff
@@ -0,0 +1,156 @@
...
+
+/**
+ * A cache of mappings from activities' identifiers (tasks, surveys, compound activities) to
+ * (SchemaReference, SurveyReference). This is to help identify how an activity's return data should
+ * be formatted.
+ */
+public class ActivitySchemaCache {
+    private static final Logger LOG = LoggerFactory.getLogger(ActivitySchemaCache.class);
+    @NonNull
+    private final Map<String, SchemaReference> taskIdentifierToSchema;
```
dephillipsmichael, line 27: I think that this data structure could be better.  5 maps that are coupled and need to be maintained together but are actually separate variables, will make it easy for someone who is not familiar with their inter-dependence to create a bug.  

I would suggest re-evaluating how this could be stored as a single Map.  My first thought would be to have a Map<String, TypeValueActivity>, where TypeValueActivity would hold the activity's type (task, survey, compound) and a getter for the Value which does a switch on the Type and casts the Object value as either SurveyReference, SchemaReference, or CompoundActivity.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/activity/ActivitySchemaCache.java
```diff
@@ -0,0 +1,156 @@
...
+        taskIdentifierToSchema = Maps.newHashMap();
+        surveyIdentifierToSurvey = Maps.newHashMap();
+        surveyGuidToIdentifier = Maps.newHashMap();
+        taskIdentifierToCompoundActivity = Maps.newHashMap();
+        identifierToActivityType = Maps.newHashMap();
+
+        addTasks();
+    }
+
+    private void addTasks() {
```
dephillipsmichael, line 48: Before I can approve this PR, let's add methods to BridgeDataProvider to set up the ActivitySchemaCache.  I think that would be the easiest time to set it up, when we create it in mPower.  Of course, we could always use a JSON file, but I think it will make more sense in code.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/activity/ActivitySchemaCache.java
```diff
@@ -0,0 +1,156 @@
...
+        taskIdentifierToSchema.put("4-APHTimedWalking-80F09109-265A-49C6-9C5D-765E49AAF5D9",
+                new SchemaReference().id("Walking Activity").revision(6L));
+    }
+
+    /**
+     * Caches the activity's schema information according to its type.
+     *
+     * @param activity activity to cache
+     */
+    public void cacheActivitySchema(Activity activity) {
```
dephillipsmichael, line 80: Now that I think about it, why don't we just store a map of <String, Activity>

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/activity/ArchiveBuilderFactory.java
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * @param identifier activity identifier (surveyGuid, taskId)
+     * @return builder for the corresponding activity, or null if not known
+     */
+    @Nullable
+    public Archive.Builder create(String identifier) {
+
+        ActivityType activityType = activityCache.getActivityType(identifier);
+
+        if (TASK == activityType) {
```
dephillipsmichael, line 43: switch statement?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/activity/ArchiveBuilderFactory.java
```diff
@@ -0,0 +1,80 @@
...
+import org.slf4j.LoggerFactory;
+
+import static org.sagebionetworks.bridge.rest.model.ActivityType.SURVEY;
+import static org.sagebionetworks.bridge.rest.model.ActivityType.TASK;
+
+/**
+ * Creates Archive.Builder instances from an activity identifier, popoulated with correct metadata
+ * to identify the format of the payload to Bridge.
+ */
+public class ArchiveBuilderFactory {
```
dephillipsmichael, line 21: I think you could move this class to the ArchiveBuild class and make an inner public static class called "Factory"

android-sdk/src/test/java/org/sagebionetworks/bridge/android/manager/activity/ArchiveBuilderFactoryTest.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Test
+    public void testMoodSurvey() {
+        Archive.Builder builder = archiveBuilderFactory
+                .create("3-APHMoodSurvey-7259AC18-D711-47A6-ADBD-6CFCECDED1DF");
+
+        Archive archive = builder.build();
+
+        assertNotNull(archive);
```
dephillipsmichael, line 50: Should we also test for the contents of the archive, like schema?

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -531,19 +534,7 @@ public Task loadTask(Context context, SchedulesAndTasksModel.TaskScheduleModel t
     public void uploadTaskResult(Context context, TaskResult taskResult) {
         // TODO: Update/Create TaskNotificationService
 
-        boolean isActivity = false;
-        if (taskResult.getTaskDetails().containsKey(ActiveTaskActivity.ACTIVITY_TASK_RESULT_KEY)) {
```
dephillipsmichael, line 535: Can you delete all references to task details then?  Seems like it is useless now.

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/TaskHelper.java
```diff
@@ -136,31 +137,50 @@ public Task loadTask(Context context, SchedulesAndTasksModel.TaskScheduleModel t
...
+        if (builder != null) {
+            uploadTaskResult(taskResult, builder);
+            return;
+        }
 
-        uploadTaskResult(taskResult, bridgeArchiveBuilder);
-    }
+        logger.debug("Handling TaskResult using TaskDetails");
+        boolean isActivity = false;
+        if (taskResult.getTaskDetails().containsKey(ActiveTaskActivity.ACTIVITY_TASK_RESULT_KEY)) {
```
dephillipsmichael, line 152: Doesn't this code go against the whole point of the activity cache?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/activity/ActivitySchemaCache.java
liujoshua: Your point about the confusing nature and complexity is a good one. I was trying to preserve the entities and denormalize, like how it is on the server side. I was not looking at the service code while writing this, but the code ended up looking quite similar. I'll think more on it, and describe more of the entities, if you have any ideas.

Activity points to one of Task or Survey or CompoundActivity
CompoundActivity has many Tasks and Surveys

Each Task/Survey/CompoundActivity can belong to many Activities

Each Task has many multiple revisions, Survey has many createdOns

liujoshua:
You've brought up many good points. I'm going to close this request and come back to this task later and focus on more pressing issues, like SignIn/Consent

We currently have something that works, and I want to give a redesign enough time.

PR #42, opened 2017-03-26
Feature/ma 140

dephillipsmichael:
Note this branch won't pass CI until https://github.com/ResearchStack/ResearchStack/pull/370 is merged and this points at RS dev.  

mPower iOS has a Medication Task that it shows before pretty much every Activity.  The medication task tracks the medication you take and based on the answers you give, will dynamically change how and when it shows you the next time.

This PR is the preliminary work on creating the Medication Task.  Of note, the task itself, and the idea of TrackedObjectTasks, are a very specific concept to mPower and Bridge, so I don't believe it belongs in ResearchStack.   This forced me to create everything here in the Bridge SDK, and so I was able to test fully customized JSON parsing into Steps and Tasks using Factory subclass'.  Most of the classes added are model classes fitting into the custom deserialization process.

### DIFF
--- a/researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/task/tracked/TrackedDataObjectCollection.java
+++ b/researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/task/tracked/TrackedDataObjectCollection.java
@@ -9,9 +9,10 @@ import org.researchstack.backbone.StorageAccess;
 import org.researchstack.backbone.result.StepResult;
 import org.researchstack.backbone.step.Step;
 import org.researchstack.backbone.storage.file.FileAccess;
-import org.researchstack.backbone.utils.LogExt;
 import org.researchstack.backbone.utils.StepResultHelper;
 import org.researchstack.backbone.utils.StepResultHelper.ResultClassComparator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.Serializable;
@@ -27,10 +28,11 @@ import java.util.Map;
 
 public class TrackedDataObjectCollection implements Serializable {
 
+    private static final Logger LOG = LoggerFactory.getLogger(TrackedDataObjectCollection.class);
     public static final String TRACKED_DATA_STORE_PATH = "TRACKED_DATA_STORE_PATH";
 
     private static final int DEFAULT_TRACKING_SURVEY_REPEAT_INTERVAL = 30 * 24 * 60 * 60;   // Every 30 days by default
-    private static final int DEFAULT_MOMENT_IN_DAY_REPEAT_INTERVAL   = 30 * 60;   // Every 20 minutes by default
+    private static final int DEFAULT_MOMENT_IN_DAY_REPEAT_INTERVAL   = 20 * 60;   // Every 20 minutes by default
 
     /**
      * Filename associated we loading and saving this tracked data object collection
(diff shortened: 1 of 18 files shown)


researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/task/tracked/TrackedDataObjectCollection.java
```diff
@@ -0,0 +1,325 @@
...
+/**
+ * Created by TheMDP on 3/21/17.
+ */
+
+public class TrackedDataObjectCollection implements Serializable {
+
+    public static final String TRACKED_DATA_STORE_PATH = "TRACKED_DATA_STORE_PATH";
+
+    private static final int DEFAULT_TRACKING_SURVEY_REPEAT_INTERVAL = 30 * 24 * 60 * 60;   // Every 30 days by default
+    private static final int DEFAULT_MOMENT_IN_DAY_REPEAT_INTERVAL   = 30 * 60;   // Every 20 minutes by default
```
liujoshua: 30 mins, or 20?

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/task/tracked/TrackedDataObjectCollection.java
```diff
@@ -0,0 +1,325 @@
...
+        return collection;
+    }
+
+    /**
+     * @return a FileAccess object, if one is available and valid, null otherwise
+     */
+    protected static FileAccess verifyFileAccess() {
+        StorageAccess storageAccess = StorageAccess.getInstance();
+        if (storageAccess == null) {
+            LogExt.e(TrackedDataObjectCollection.class, "StorageAccess is null");
```
liujoshua: I'd prefer that Log4j is used instead of LogExt in the Bridge packages, so the format of logging and log levels are controlled the same way between android-sdk and researchstack-sdk. Thoughts?

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/task/tracked/TrackedDataSelectedItemsInterface.java
```diff
@@ -0,0 +1,17 @@
...
+
+import org.researchstack.backbone.result.StepResult;
+
+import java.util.List;
+
+/**
+ * Created by TheMDP on 3/23/17.
+ */
+
+public interface TrackedDataSelectedItemsInterface {
```
liujoshua, line 11: Would it make sense to call this "TrackedDataStep" or something? Remove "Interface" from the name. Remove the concept of "selected", since this could work for any collection of TrackedDataObjects, and in this case, the caller gets the "selected" item results, because they pass in the "selected" objects

liujoshua (approved):
Looks good, minor suggestions

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/task/tracked/TrackedDataObjectCollection.java
dephillipsmichael: Good point, i meant 20 mins, will switch the constant

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/task/tracked/TrackedDataObjectCollection.java
dephillipsmichael: No preference, as long as we're consistent.

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/task/tracked/TrackedDataSelectedItemsInterface.java
dephillipsmichael: Those are good thoughts, and I think you are on to something. I'm not 100% this class is even needed.  It was created as I converted from iOS, and need to match parity with it's use of multiple inheritance. I'm going to put this on my radar, and avoid refactoring it just yet,  and as I finish the TrackedDataTask and it's unit tests, I will re-evaluate what this should be called and how it should operate.

liujoshua: Sounds good, feel free to merge whenever you're ready. I've already approved this PR since everything looks greatand makes sense as is

PR #43, opened 2017-03-29
bug/data_sharing

dephillipsmichael:
fixed crash setting user sharing scope

### DIFF
--- a/researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
+++ b/researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
@@ -484,8 +484,6 @@ public abstract class BridgeDataProvider extends DataProvider {
         SharingScope sharingScope = RestUtils.GSON.fromJson(scope, SharingScope.class);
         participant.setSharingScope(sharingScope);
 
-        participant.setSharingScope(SharingScope.valueOf(scope));
-
         setUserSharingScope(sharingScope)
                 .subscribe();
     }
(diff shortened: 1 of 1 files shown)


researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -480,19 +481,22 @@ public SharingScope getUserSharingScope() {
...
     public void setUserSharingScope(Context context, String scope) {
         StudyParticipant participant = new StudyParticipant();
+        SharingScope sharingScope = RestUtils.GSON.fromJson(scope, SharingScope.class);
+        participant.setSharingScope(sharingScope);
+
         participant.setSharingScope(SharingScope.valueOf(scope));
 
-        setUserSharingScope(scope)
+        setUserSharingScope(sharingScope)
                 .subscribe();
```
liujoshua: do you think subscribe() is enough here, or should this block until the call is done?

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -480,19 +481,22 @@ public SharingScope getUserSharingScope() {
     @Override
     public void setUserSharingScope(Context context, String scope) {
         StudyParticipant participant = new StudyParticipant();
+        SharingScope sharingScope = RestUtils.GSON.fromJson(scope, SharingScope.class);
+        participant.setSharingScope(sharingScope);
+
         participant.setSharingScope(SharingScope.valueOf(scope));
```
liujoshua: Is this line supposed to be deleted?

liujoshua (commented):
Thanks for fixing

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
dephillipsmichael: I don't think it's a good idea to have a public void method in DataProvider block whatever thread it's on

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
dephillipsmichael: Yes it is, good find.  I must have been debugging something before I pushed the code.

liujoshua (approved):

PR #49, opened 2017-10-14
Bring in RSuite 0.0.3; Test that Room works

liujoshua:


EricSiegNW:
Is it correct to be adding ResearchSuite stuff to BridgeAndroidSDK?

liujoshua:
Currently, BridgeAndroidSDK has two modules: a Bridge client/SDK for Android and a ResearchStack SDK. ResearchSuite is different enough it'll allow a more conventional Android architecture (i.e not needing to implement DataProvider) and I intend to put that code here. That way the ResearchSuite repo won't have Bridge specific code. In some ways, ResearchStack was too tied to Bridge, so some of that functionality will move here.

The Bridge Research SDK will stay intact so we can still support apps coding against ResearchStack (like mPower). We want to support migrating between ResearchKit/Stack and ResearchSuite, so keeping these two available will help us understand how well we're meeting that goal.

For ResearchSuite, we'll refactor/build it according to priorities for new deliverables.

Do you have thoughts on where this code should go?

researchsuite-sdk/build.gradle
```diff
@@ -0,0 +1,46 @@
...
+            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+        }
+    }
+}
+
+dependencies {
+    compile fileTree(dir: 'libs', include: ['*.jar'])
+    compile 'com.android.support:appcompat-v7:26.1.0'
+
+    compile 'org.sagebionetworks.researchsuite:api:0.0.3'
```
dephillipsmichael, line 35: Where can I find documentation on this library?

researchsuite-sdk/build.gradle
```diff
@@ -0,0 +1,46 @@
...
+
+    compile "android.arch.persistence.room:runtime:1.0.0-beta2"
```
dephillipsmichael: Room seems promising.  How well does it integrate/sync with the network when online/offline etc?  Will that be the responsibility of ResearchSuite, or will that somehow be tied to BridgeAndroidSDK?

liujoshua: I'd like uploads and scheduled activities to use this model:
- ResearchSuite talks to to sqlite (putting this in BridgeSDK repo for now, but ideally it would be general enough to move to to ResearchSuite)
- [SyncAdapters](https://developer.android.com/training/sync-adapters/index.html) do the heavy lifting of syncing DB to REST API. This would live in BridgeSDK. The initial implementation of the SyncAdapters would call into the existing managers.

researchsuite-sdk/build.gradle
liujoshua: Here's a (outdated) diagram here for Android: https://github.com/Sage-Bionetworks/ResearchSuite-Android/pull/1

Shannon and I are working on documenting the high level design. Shannon would be the best resource on what the end result/goals look like. For now it'd be good to follow pull requests here as we flesh things out:
https://github.com/Sage-Bionetworks/ResearchSuite-Android/
https://github.com/Sage-Bionetworks/ResearchSuite-AppleOS/

PR #58, opened 2017-10-27
AA-50 crash fix; AA-9 crash fix

liujoshua:
* Use AtomicReference to always hold latest authenticated API
* Loosen consent signature field nullity requirements to allow
  local storage of partially completed signatures


android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
```diff
@@ -130,11 +141,15 @@ public Completable signUp(@NonNull final SignUp signUp) {
...
-                                    .password(signUp.getPassword()));
+                    SignIn signIn = new SignIn()
+                            .study(config.getStudyId())
+                            .email(signUp.getEmail())
+                            .password(signUp.getPassword());
+
+                    accountDAO.setSignIn(signIn);
+                    forConsentedUsersApiAtomicReference.set(
+                            apiClientProvider.getClient(ForConsentedUsersApi.class, signIn)
+                    );
```
DwayneJengSage, line 152: Should we update the signUp() tests to verify that we're updating the apiAtomicReference?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
```diff
@@ -208,6 +226,9 @@ public Completable resendEmailVerification(@NonNull final String email) {
                 }).doOnError(throwable -> {
                     // a 412 is a successful signin
                     if (throwable instanceof ConsentRequiredException) {
+                        forConsentedUsersApiAtomicReference.set(
+                                apiClientProvider.getClient(ForConsentedUsersApi.class, signIn)
+                        );
```
DwayneJengSage, line 231: Do we also want to update AuthenticationManagerTest.signIn_UploadLocalConsentAndRetrySignIn() to test for this?

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -167,63 +181,90 @@ public boolean isConsented() {
...
+        return consent;
     }
 
     @Override
-    public void saveConsent(Context context, TaskResult consentResult) {
-        giveConsent(bridgeConfig.getStudyId(), createConsentSignature(consentResult));
+    public void saveConsent(Context context, @NonNull TaskResult consentResult) {
+        logger.warn("saveConsent called -- DEPRECATED");
+        giveConsent(bridgeConfig.getStudyId(), createConsentSignature(consentResult))
+                .toBlocking().value();
```
DwayneJengSage: Is .toBlocking().value() supposed to be here? Or is that some debugging code that got left behind?

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -167,63 +181,90 @@ public boolean isConsented() {
...
+        logger.debug("loadLocalConsent called, got: ");
+        return consent;
     }
 
     @Override
-    public void saveConsent(Context context, TaskResult consentResult) {
-        giveConsent(bridgeConfig.getStudyId(), createConsentSignature(consentResult));
+    public void saveConsent(Context context, @NonNull TaskResult consentResult) {
+        logger.warn("saveConsent called -- DEPRECATED");
+        giveConsent(bridgeConfig.getStudyId(), createConsentSignature(consentResult))
```
DwayneJengSage: My intuition tells me that whatever bug you were trying to fix in saveLocalConsent() is likely to manifest here as well. Would it be better to have saveConsent() call saveLocalConsent(), and then call the server to upload consent? (Or alternatively, refactor it into a helper method that both saveConsent() and saveLocalConsent() call?)

researchstack-sdk/src/test/java/org/sagebionetworks/bridge/researchstack/BridgeDataProviderTest.java
```diff
@@ -237,6 +240,13 @@ public void testSaveConsent() {
     }
 
     @Test
+    public void saveLocalConsent_scopeOnly() throws Exception {
+        ConsentSignatureBody body = new ConsentSignatureBody();
+        body.scope = SharingScope.ALL_QUALIFIED_RESEARCHERS.toString();
+        dataProvider.saveLocalConsent(context, body);
+    }
```
DwayneJengSage: This tests that saveLocalConsent() doesn't crash, but we should also verify that authenticationManager is being called with the correct args.

We should also have a separate test case with _all_ the fields.

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -167,63 +181,90 @@ public boolean isConsented() {
...
+                toSharingScope(signatureBody.scope));
+    }
+
+    private SharingScope toSharingScope(String sharingScope) {
+        SharingScope scopeEnum = SharingScope.NO_SHARING;
+        for (SharingScope scope : SharingScope.values()) {
+            if (scope.toString().equals(sharingScope)) {
+                scopeEnum = scope;
+            }
+        }
```
DwayneJengSage, line 210: Would this code be cleaner with SharingScope.valueOf()? (Note that valueOf() throws NullPointerException and IllegalArgumentException, so you may have to handle those.)

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -167,63 +181,90 @@ public boolean isConsented() {
...
+        return scopeEnum;
     }
 
     @Nullable
-    protected ConsentSignature createConsentSignature(ConsentSignatureBody consentSignatureBody) {
+    protected ConsentSignature createConsentSignature(@Nullable ConsentSignatureBody
+                                                              consentSignatureBody) {
```
DwayneJengSage: This is called by uploadConsent() (which has a test in BridgeDataProviderTest that's currently marked as Ignore). Should we un-ignore that test and update it? We should have two separate test cases, one with all fields specified, and one with only required fields specified.

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -167,63 +181,90 @@ public boolean isConsented() {
...
-            }
-        }
         signature.setScope(sharingScope);
         return signature;
     }
 
     @Nullable
-    protected ConsentSignatureBody createConsentSignatureBody(ConsentSignature consentSignature) {
+    protected ConsentSignatureBody createConsentSignatureBody(@Nullable ConsentSignature
+                                                                      consentSignature) {
```
DwayneJengSage: Similarly, should we add unit tests for this? (Called by loadLocalConsent().) Again, we should have two test cases, one for all fields, one for only required fields.

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -167,63 +181,90 @@ public boolean isConsented() {
...
     }
 
     @Override
     public void saveLocalConsent(Context context, ConsentSignatureBody signatureBody) {
-        giveConsent(bridgeConfig.getStudyId(), createConsentSignature(signatureBody));
+        logger.debug("saveLocalConsent called for: " + signatureBody);
+        LocalDate birthdate = null;
+        if (signatureBody.birthdate != null) {
+            birthdate = LocalDate.fromDateFields(signatureBody.birthdate);
+        }
```
DwayneJengSage: Would it be useful to have a wrapper function that takes in a ConsentSignature and calls authenticationManager.giveConsentSync()? This way, you can leverage the null checks and type conversions of createConsentSignature() and avoid writing these checks in two different placse.

DwayneJengSage (changes requested):
Mostly notes about additional unit tests that I think we need. The tests should be relatively straightforward to add.

A few comments about refactor saveConsent() and saveLocalConsent().

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
liujoshua: Updated

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
liujoshua: My opinion is that it's better to have signIn() test for all primary responsibilities of successful signing in, while this tests tests that a signIn followed by 412 will trigger an upload (while signIn_UploadLocalConsentRetrySignInOnce checks that there is only one attempt to upload before propagating the exception). 

Still not following "one assertion per test", but trying to follow the spirit that each test focuses on one aspect of functionality.

researchstack-sdk/src/test/java/org/sagebionetworks/bridge/researchstack/BridgeDataProviderTest.java
liujoshua: Agreed

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
liujoshua: This was incorrect

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
liujoshua: In this case, the enum's toString() and name() are different. Verified via unit test valueOf does not work

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
liujoshua: I did some refactoring and think I did what you suggested? Not entirely sure what you envisioned.

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
liujoshua: This path is deprecated and unused by all our code, will throw UnsupportedOperation() instead of having untested code.

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
liujoshua: Wrote a test that covers this

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
DwayneJengSage, line 231: That's fine, but we need at least one test case that covers this. Whether we update the existing test or create a new test, that's immaterial.

DwayneJengSage (changes requested):
One comment about missing unit test. Otherwise, looks good.

DwayneJengSage (approved):

PR #59, opened 2017-10-31
add Task ID to Schema mapping

DwayneJengSage:
Currently, BridgeAndroidSDK assumes that all schema revs are 1. This fix adds a static JSON mapping, read from the app's assets folder, to map task IDs to schema IDs/revs.

Also removed the default public key, as per conversation with Josh.

Testing done:
* added unit tests for changed files
* ./gradlew check
* implemented in mPower Android and manually tested


android-sdk/src/main/java/org/sagebionetworks/bridge/android/BridgeConfig.java
```diff
@@ -104,15 +137,24 @@ public String getAppVersionName() {
 
     @NonNull
     public X509Certificate getPublicKey() throws IOException, CertificateException {
-        InputStream publicKeyFile;
-        try {
-            publicKeyFile = applicationContext.getAssets().open(STUDY_PUBLIC_KEY, ACCESS_BUFFER);
+        try (InputStream publicKeyFile = applicationContext.getAssets().open(STUDY_PUBLIC_KEY,
```
liujoshua, line 140: Thanks for finding this

liujoshua (approved):

PR #63, opened 2017-11-04
Activity functionality

dephillipsmichael:
These are needed to set up the scheduling on CRF, see my branch https://github.com/dephillipsmichael/CardiorespiratoryFitness-Android/tree/new_scheduling for how this will be used in CRF


DwayneJengSage (approved):
Implementation looks good. A few questions about testing requirements:

* I assume these changes were integrated with and tested in your local CRF workspace.
* Do we have any unit test requirements? All these methods are passthroughs with no business logic, so they're not particularly interesting to test. On the other hand, it's not difficult to test them with mocks to verify that we're passing back the correct args and calling the correct APIs. Should we require unit tests for passthroughs or are we fine waiving them?

liujoshua (approved):

dephillipsmichael:
IMO unit tests for pass throughs provide no value. However, I wrote a bunch of tests for the business logic in the CrfDataProviderTest that use these methods in a complex way

PR #64, opened 2017-11-04
support server-side surveys

DwayneJengSage:
Added SurveyManager and updated BridgeDataProvider and TaskHelper to enable loading server-side surveys. Added factories to facilitate unit tests.

Depends on https://github.com/ResearchStack/ResearchStack/pull/401

Testing done:
* ./gradlew check
* Tested through mPower-Android and CRF Module


researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/factory/TaskFactory.java
```diff
@@ -0,0 +1,35 @@
...
+import android.content.Context;
+import android.support.annotation.NonNull;
+import org.researchstack.skin.model.TaskModel;
+import org.researchstack.skin.task.SmartSurveyTask;
+
+/**
+ * This class encapsulates creating Task instances, to allow us to separate the logic of creating
+ * Tasks from using tasks. This helps simplify tests.
+ */
+public class TaskFactory {
```
dephillipsmichael: Did you look at SurveyFactory or TaskItemFactory? I'd suggest just adding a method in SurveyFactory for now with newSmartSurveyTask and using that class.

dephillipsmichael (changes requested):
Nice work. Everything looks good besides the duplicate TaskFactory, please remove that and add smart survey functionality to SurveyFactory or TaskItemFactory.  This will include moving SmartSurveyTask and TaskModel down to backbone as well.

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -580,22 +577,20 @@ private SchedulesAndTasksModel translateActivities(@NonNull ScheduledActivityLis
...
+        List<Integer> dayList = new ArrayList<>(activityMap.keySet());
+        Collections.sort(dayList);
+
+        // Convert from Bridge activities to ResearchKit task model
         SchedulesAndTasksModel model = new SchedulesAndTasksModel();
         model.schedules = new ArrayList<>();
-        for (int day : activityMap.keySet()) {
+        for (int day : dayList) {
             List<ScheduledActivity> aList = activityMap.get(day);
             ScheduledActivity temp = aList.get(0);
```
dephillipsmichael, line 596: This isn't related to your PR, but I just found that the code in this method has a critical error regarding daylight saving time:
"sm.scheduledOn = temp.getScheduledOn().toDate();"
Will translate the scheduledOn date incorrectly if today is before daylight, and the activity is scheduled after daylight. It should be...
"sm.scheduledOn = temp.getScheduledOn().toLocalDate().toDate();"

DwayneJengSage: What's the semantics of ScheduleModel.scheduledOn? This is defined as a java.util.Date, which appears to be a timestamp, not a calendar date. Is it correct to flatten this to midnight?

liujoshua (approved):

PR #66, opened 2017-11-10
Bugfix/aa 54 missing survey answers

liujoshua:


researchstack-sdk/src/test/java/org/sagebionetworks/bridge/researchstack/factory/ArchiveFileFactoryTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+        SurveyAnswer.ChoiceSurveyAnswer deserializedAnswer =
+                RestUtils.GSON.fromJson(jsonString, SurveyAnswer.ChoiceSurveyAnswer.class);
+
+        assertEquals(answer.item, deserializedAnswer.item);
+        assertEquals(answer.startDate, deserializedAnswer.startDate);
+        assertEquals(answer.endDate, deserializedAnswer.endDate);
+        assertEquals(answer.questionType, deserializedAnswer.questionType);
+        assertEquals(answer.questionTypeName, deserializedAnswer.questionTypeName);
+        assertEquals(answer.choiceAnswers, deserializedAnswer.choiceAnswers);
```
DwayneJengSage, line 58: Can we explicitly test that deserializedAnswer.choiceAnswers contains "Result1". I'm concerned that if there's somehow another bug preventing the answer from being wired through, we may end up just checking that null equals null in this test.

DwayneJengSage (approved):
Looks good. Just one comment about making the unit test more explicit about what we're testing for.

DwayneJengSage (approved):

PR #69, opened 2017-11-17
data groups in onboarding

DwayneJengSage:
Add the data group step to onboarding. See https://sagebionetworks.jira.com/browse/AA-78

Testing done:
* ./gradlew check
* Manually tested both test_user and participant branches.
* Set a breakpoint to verify we only save data groups at the last step, rather than a separate server call each step.


dephillipsmichael:
Your test coverage is immaculate, nice work.  Approved :shipit:

dephillipsmichael (approved):

PR #79, opened 2017-12-01
bug/skip_crash

dephillipsmichael:
Fixed bug with skipped question results (specifically choice answers)

### DIFF
--- a/researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/factory/ArchiveFileFactory.java
+++ b/researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/factory/ArchiveFileFactory.java
@@ -69,7 +69,10 @@ public class ArchiveFileFactory {
         checkNotNull(result.getEndDate());
 
         if (result instanceof StepResult) {
-            return fromStepResult((StepResult) result);
+            // Skipped steps will have a null result
+            if (((StepResult) result).getResult() != null) {
+                return fromStepResult((StepResult) result);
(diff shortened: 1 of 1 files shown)


researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/factory/ArchiveFileFactory.java
```diff
@@ -69,7 +69,10 @@ public ArchiveFile fromResult(@NonNull Result result) {
         checkNotNull(result.getEndDate());
 
         if (result instanceof StepResult) {
-            return fromStepResult((StepResult) result);
+            // Skipped steps will have a null result
```
liujoshua: Can we do this null handling in fromStepResult? 
We'd still need to check return value of fromStepResult to know whether to return early, but new code that calls fromStepResult without null checking first would no longer crash

dephillipsmichael:
@liujoshua random issue, but see the CI, it looks like the 3rd party graph dependency may have been deleted.  Ideas on resolution?

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/factory/ArchiveFileFactory.java
dephillipsmichael: Sure, I'll move it there.

liujoshua (approved):
Looks good. One suggestion

PR #80, opened 2017-12-02
Bug/AA-8, AA-9, AA-125

dephillipsmichael:
This address:
https://sagebionetworks.jira.com/browse/AA-8
https://sagebionetworks.jira.com/browse/AA-9
https://sagebionetworks.jira.com/browse/AA-125

### DIFF
--- a/android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
+++ b/android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
@@ -38,6 +38,7 @@ import java.util.concurrent.atomic.AtomicReference;
 import rx.Completable;
 import rx.Observable;
 import rx.Single;
+import rx.functions.Action1;
 import rx.functions.Func1;
 
 import static com.google.common.base.Preconditions.checkNotNull;
@@ -263,7 +264,7 @@ public class AuthenticationManager {
 
                     if (!(throwable instanceof ConsentRequiredException)
                             || retryAttempt > 1
-                            || !isConsented()) {
+                            || isConsentedOnBridgeServer()) {
                         Observable<UserSessionInfo> obs = Observable.error(throwable);
                         return obs;
                     }
@@ -401,19 +402,6 @@ public class AuthenticationManager {
         listeners.remove(listener);
     }
 
-
-    // region Consent
-
-    /**
-     * @param subpopulationGuid guid for the subpopulation of the consent
-     * @return true if consent to participate was made for this consent
-     */
-    public boolean isConsented(@NonNull String subpopulationGuid) {
-        checkNotNull(subpopulationGuid);
-
-        return isConsentedInSessionOrLocal(getUserSessionInfo(), subpopulationGuid);
-    }
-
     /**
      * @param subpopulationGuid guid for the subpopulation of the consent
      * @return true if the consent to participate was made against the most recently published
@@ -440,18 +428,31 @@ public class AuthenticationManager {
 
     // if the participant's session indicates consent to this subpopulation, use that. otherwise,
     // treat presense of consent in DAO as having consented
-    boolean isConsentedInSessionOrLocal(UserSessionInfo session, String subpopulationGuid) {
+    boolean isConsented(UserSessionInfo session, String subpopulationGuid, boolean checkLocalCache) {
         ConsentStatus subpopulationStatus = getConsentStatusFromSession(subpopulationGuid);
         if (subpopulationStatus != null && subpopulationStatus.getConsented()) {
             return true;
         }
-        return consentDAO.getConsent(subpopulationGuid) != null;
+        if (checkLocalCache) {
+            return consentDAO.getConsent(subpopulationGuid) != null;
+        }
+        return false;
+    }
+
+    public boolean isConsentedOnBridgeServer() {
+        return isConsented(false);
+    }
+
+    public boolean isConsentedOnBridgeOrLocally() {
+        return isConsented(true);
     }
 
     /**
+     * @param checkLocalCache if true, we will check if the user has consented locally,
+     *                        if false, we will check if consent has been verified on the server ONLY
      * @return true if all required consents have been signed
      */
-    public boolean isConsented() {
+    private boolean isConsented(boolean checkLocalCache) {
         UserSessionInfo userSessionInfo = getUserSessionInfo();
         if (userSessionInfo != null) {
             if (userSessionInfo.getConsented()) {
@@ -459,7 +460,7 @@ public class AuthenticationManager {
             }
 
             for (String subpopulation : getRequiredConsents(userSessionInfo)) {
-                if (!isConsentedInSessionOrLocal(userSessionInfo, subpopulation)) {
+                if (!isConsented(userSessionInfo, subpopulation, checkLocalCache)) {
                     return false;
                 }
             }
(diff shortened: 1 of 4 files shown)


android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
```diff
@@ -571,9 +574,6 @@ public ConsentSignature storeLocalConsent(@NonNull String subpopulationGuid, @No
                                               @Nullable String imageMimeType,
                                               @NonNull SharingScope sharingScope) {
         checkNotNull(subpopulationGuid);
-        checkNotNull(name);
```
dephillipsmichael: Removing these fixes AA-9, so we can save incomplete consent signatures

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
```diff
@@ -263,7 +264,7 @@ public Completable resendEmailVerification(@NonNull final String email) {
 
                     if (!(throwable instanceof ConsentRequiredException)
                             || retryAttempt > 1
-                            || !isConsented()) {
+                            || isConsentedOnBridgeServer()) {
```
dephillipsmichael: These changes fix AA-8. There were two bugs here, !isConsented() was an incorrect conditional that was causing the error observable to always be returned and uploadConsentLocally() never was attempted.  And, the method "isConsented()" was inappropriate here since that call looks at local consent, and would always be considered consented in contrast to the ConsentRequiredException.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
```diff
@@ -401,19 +402,6 @@ public void removeEventListener(AuthenticationEventListener listener) {
...
     }
 
-
-    // region Consent
-
-    /**
-     * @param subpopulationGuid guid for the subpopulation of the consent
-     * @return true if consent to participate was made for this consent
-     */
-    public boolean isConsented(@NonNull String subpopulationGuid) {
```
dephillipsmichael: Never used, removed

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -386,7 +386,8 @@ public Completable signIn(@NonNull String email, @NonNull String password) {
     }
 
     public boolean isSignedIn() {
-        return authenticationManager.getEmail() != null;
+        return authenticationManager.getEmail() != null &&
+                authenticationManager.isConsentedOnBridgeServer();
```
dephillipsmichael: This fixes AA-125 by adding a check for consent on the bridge (through the web service returned UserSession) that if not hit, will push the user to the email verification screen

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
liujoshua: I'm pretty sure it the previous version is correct.

If there is no local consent, there is no reason to attempt another upload, and the ConsentRequiredException should be  propagated. I think that the onError in the Observerable's subscriber is not implemented, which is the real issue. onError should be triggering the consent flow. 

isConsentedOnBridgeServer check is fine, but should be redundant. isConsentedOnBridgeServer only runs if throwable was a ConsentRequiredException, in which case isConsentedOnBridgeServer should always be false. If that's not the case, there's a bug we need to fix.

I don't understand how this fixes AA-8. But at the same time, this change shouldn't hurt anything (other than maybe attempt an unnecessary upload), so if it is somehow fixing AA-8 I'm okay leaving it.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
liujoshua: Odd, I remember doing a PR on these with Dwayne. Maybe the PR was closed.

Can you update the `@NonNull` to `@Nullable`?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
```diff
@@ -597,7 +597,6 @@ public ConsentSignature storeLocalConsent(@NonNull String subpopulationGuid, @No
     public void storeLocalConsent(@NonNull String subpopulationGuid,
                                   @NonNull ConsentSignature consentSignature) {
```
liujoshua: Can you update the `@NonNull` to `@Nullable`?

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
liujoshua: I'm not sure this is the correct way to address AA-125. You can be email verified without being consented.

How about authenticationManager.getUserSessionInfo() != null?

liujoshua: IIRC from iOS discussions, the way to detect email verification is the correct next step, is to know the user has signed up, yet the email/password used to sign in failed. 

I know there's been updates to Auth, so I'll look into whether we now provide a real API to know sign in failed due to lack of email verification, and not incorrect email/password

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
liujoshua: Was saving of partial consents in AA-9 not fixed? It should've been fixed here: https://github.com/Sage-Bionetworks/BridgeAndroidSDK/pull/58/files

storeLocalConsent(String, ConsentSignature) is the one for saving incomplete consent signatures. This method is supposed to store completed ones. I'd argue part of the confusion is we should have two separate stores -- for completed and incompleted consents. Storing completed consents is a cache for uploading. Storing incompleted consents is for saving data from various steps towards building a completed consent.

IIRC the remaining work for AA-9 was not in this API, but something in the steps wasn't properly ever attempting to save the sharing scope.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
dephillipsmichael: Your logic makes sense to me when reading it, it may be possible that the changes I made were unnecessary.  I wasn't thinking in terms of "If there is no local consent, there is no reason to attempt another upload", I'll add that comment to "isConsented()", revert those methods changes, and see if it works.

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
dephillipsmichael: I thought the call to sign up created a UserSession, but it doesn't look like it does.  You are correct then, that would be a better test for if email is verified.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
```diff
@@ -199,8 +199,10 @@ public Completable resendEmailVerification(@NonNull final String email) {
 
         logger.debug("signIn called with email: " + email);
 
+        String subpopulationGuid = config.getStudyId();
```
liujoshua, line 202: Nit-pick: conceptually this makes more sense as studyId.

We programatically create a default consent with subpopulationGuid = studyId, so passing in a studyId as a subpopulationGuid makes sense in this context, but passing a subpopulationGuid (which is more generally a GUID/UUID) into a study is confusing. 

OHSUMoleMapper is a study that has multiple consents, idk of an example for an iOS study we could look at

liujoshua (approved):

PR #88, opened 2017-12-10
Improve dataGroup handling

liujoshua:
* Allow skipping of dataGroup step if dataGroup already set
* Update local dataGroup store during StudyParticipant operations


researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -540,17 +540,25 @@ public void setUserSharingScope(Context context, String scope) {
         return bridgeManagerProvider.getParticipantManager()
                 .updateParticipantRecord((StudyParticipant) new StudyParticipant()
                         .email(authenticationManager.getEmail())
-                        .sharingScope(scope));
+                        .sharingScope(scope))
+                .doOnSuccess(session -> bridgeManagerProvider.getAccountDao()
```
dephillipsmichael, line 544: All these setDataGroups spots seem to stem from the user session, maybe the call to "setUserSession" should go through and update the study participant, data groups, and session?

dephillipsmichael (approved):
Looks good to me, nice use of the skip nav feature

PR #99, opened 2017-12-21
feature/editProfile

dephillipsmichael:
Added download data public function

### DIFF
--- a/researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
+++ b/researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
@@ -561,6 +561,24 @@ public abstract class BridgeDataProvider extends DataProvider {
                 .toObservable();
     }
 
+    /**
+     * Make participant data available for download.
+     * <p>
+     * Request the uploaded data for this user, in a given time range (inclusive). Bridge will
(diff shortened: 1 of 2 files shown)


researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -561,6 +561,24 @@ public void setUserSharingScope(Context context, String scope) {
...
+     * Request the uploaded data for this user, in a given time range (inclusive). Bridge will
+     * asynchronously gather the user's data for the given time range and email a secure link to the
+     * participant's registered email address.
+     *
+     * @param startDate The first day to include in reports that are returned (required)
+     * @param endDate   The last day to include in reports that are returned (required)
+     * @return completable
+     */
+    @NonNull
+    public Observable<Message> downloadData(LocalDate startDate,
```
liujoshua: The data provider generally returns only DataResponse objects

liujoshua (approved):

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
dephillipsmichael: Oh right, my fault.  I'll change that.

PR #115, opened 2018-01-23
Feature/email only auth

liujoshua:


android-sdk/src/main/java/org/sagebionetworks/bridge/android/AuthManagementActivity.java
```diff
@@ -0,0 +1,80 @@
...
+                    Intent launchIntentForPackage = pm.getLaunchIntentForPackage(getPackageName());
+                    startActivity(launchIntentForPackage);
+                }, t -> {
+                    logger.warn("Failed to authenticated: ", t);
+                    setResult(RESULT_CANCELED);
+                });
+
+        data.getLastPathSegment();
+    }
+    void verifyEmail(Uri data) {
```
dephillipsmichael: Is there a scenario where this is called?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/AuthManagementActivity.java
```diff
@@ -0,0 +1,80 @@
...
+                startSession(data);
+                break;
+            case "verifyEmail.html":
+                startSession(data);
+                break;
+            default:
+                startSession(data);
+        }
+
+        BridgeManagerProvider.getInstance()
```
dephillipsmichael: Seems like signInViaEmailLink may get called twice... here, and in startSession in the switch statement above, no?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
```diff
@@ -62,55 +66,118 @@
...
+            ApiClientProvider.AuthenticatedClientProviderBuilder builder =
+                    apiClientProvider
+                            .getAuthenticatedClientProviderBuilder()
+                            .withEmail(email);
+
+            boolean hasPasswordOrSession = false;
+            String password = accountDAO.getPassword();
+            if (!Strings.isNullOrEmpty(password)) {
+                hasPasswordOrSession = true;
+                builder.withPassword(email);
```
dephillipsmichael: should this be builder.withPassword(password)?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/AuthManagementActivity.java
liujoshua: No, this isn't needed anymore. The email verification will happen automatically on Bridge with passwordless auth flow

android-sdk/src/main/java/org/sagebionetworks/bridge/android/AuthManagementActivity.java
liujoshua: Yes, this code was moved to startSession but wasn't removed from here

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
liujoshua: Yes

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
```diff
@@ -211,34 +342,73 @@ public Completable resendEmailVerification(@NonNull final String email) {
...
+                            ConsentStatus consentStatus = consentStatusEntry.getValue();
+
+                            // required consent missing on Bridge and present locally
+                            if (consentStatus.getRequired()
+                                    && !consentStatus.getConsented()
+                                    && isConsentedInLocal(subpopulationGuid)) {
+
+                                // upload local consents, ignoring errors
+                                uploadLocalConsents()
+                                        .onErrorResumeNext(Observable.just(session))
```
dephillipsmichael, line 399: Do these errors get caught and sent to this part?...

.doOnError(t -> {
 +                    accountDAO.setSignIn(null);
 +                    accountDAO.setPassword(null);
 +                    accountDAO.setUserSessionInfo(null);
 +
                  });

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/EmailSignInRedirectUriActivity.java
```diff
@@ -0,0 +1,25 @@
...
+import org.slf4j.LoggerFactory;
+
+public class EmailSignInRedirectUriActivity extends AppCompatActivity {
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        Intent launchIntent = new Intent(this, AuthManagementActivity.class)
+                .setData(getIntent().getData())
+        .addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP | Intent.FLAG_ACTIVITY_SINGLE_TOP);
```
dephillipsmichael, line 20: Was this pass-through activity added for the sole purpose of adding Intent.FLAG_ACTIVITY_CLEAR_TOP | Intent.FLAG_ACTIVITY_SINGLE_TOP?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
liujoshua: If you're asking about uploadLocalConsents(), no. The onErrorResumeNext will just return the same session

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/EmailSignInRedirectUriActivity.java
liujoshua: Good point, this actually isn't needed because we're not using a Chrome Tab at any point. Pretty sure I included this to destroy a Chrome Tab from the backstack

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
dephillipsmichael: For all intents and purposes, wouldn't that be considered a fail on the original network request "sign in"?

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/EmailSignInRedirectUriActivity.java
dephillipsmichael: Ohhhh I ran into that too when doing fit bit auth, it was really annoying. The Chrome tab kept sticking around in all sorts of odd scenarios.  Thanks for the background info

liujoshua:
Reopening temporarily for travis

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
liujoshua: I did it this way so the consent status is "invisible" when . When we call sign in and the participant is not consented, we will attempt to upload, but as the caller of sign in, you don't have to know whether this upload call was needed, or whether upload succeeded or failed. You just need to handle errors related to not being able to authenticate (obtain a UserSessionInfo). This is also why above this lambda, I have an onErrorResumeNext which will convert ConsentRequiredException into a successful sign in and give you the UserSessionInfo.

PR #116, opened 2018-01-24
Feature/aa 248 data group step skip logic

liujoshua:


researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/step/DataGroupQuestionStep.java
```diff
@@ -35,7 +40,7 @@
 public class DataGroupQuestionStep extends NavigationExpectedAnswerQuestionStep
         implements NavigableOrderedTask.NavigationSkipRule {
     private boolean shouldPersist = false;
-    private boolean shouldSkipIfSessionContainsDataGroups = false;
```
dephillipsmichael, line 38: Any harm in leaving this functionality around?  I think we're using it in BP

dephillipsmichael (approved):
minor comment looks good

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/step/DataGroupQuestionStep.java
liujoshua: I think just the new functionality is better. The dataGroups have to be whitelisted on the server side, and then all valid dataGroups could be added to reproduce the old behavior.. This would be more work for simple scenarios, but once there are multiple steps for setting dataGroups, this was by far the cleanest solution of several I tried. Having to handle a skipIfAny in the general case was really messy

PR #127, opened 2018-02-05
bug/requiredConsentWithSignIn

dephillipsmichael:
Fixed issue with sign in succeeding before user consent

### DIFF
--- a/android-sdk/src/main/AndroidManifest.xml
+++ b/android-sdk/src/main/AndroidManifest.xml
@@ -8,23 +8,24 @@
     <application
         android:label="@string/app_name"
         android:supportsRtl="true">
-        <activity
-            android:name="org.sagebionetworks.bridge.android.AuthManagementActivity"
-            android:launchMode="singleTask">
-            <tools:validation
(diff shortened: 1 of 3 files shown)


android-sdk/src/main/AndroidManifest.xml
```diff
@@ -8,23 +8,24 @@
     <application
         android:label="@string/app_name"
         android:supportsRtl="true">
-        <activity
-            android:name="org.sagebionetworks.bridge.android.AuthManagementActivity"
-            android:launchMode="singleTask">
-            <tools:validation
-                testUrl="https://webservices.sagebridge.org/mobile/startSession.html?email=test1%40sagebase.org&amp;study=study-id&amp;token=TOKEN"/>
+        <!-- Implement this activity for email-only sign in -->
```
dephillipsmichael, line 11: By defining this in the library's manifest it makes it impossible to override.  Usually you let the app define, and put it in the sdk docs that you need to.

liujoshua (approved):

PR #132, opened 2018-02-12
Add steth, with a resource value to enable/disable

liujoshua:


android-sdk/build.gradle
```diff
@@ -55,6 +55,12 @@ dependencies {
     // marked api due to propagation of CMSException
     api 'com.madgag.spongycastle:bcpkix-jdk15on:1.58.0.0'
 
+    // Debug library for
+    implementation 'com.facebook.stetho:stetho:1.5.0'
```
dephillipsmichael, line 59: If you do intend to add Stetho to the bridge sdk, you need to conditional compile these libraries in.  As it is now, it will always include that code, which is not what we want.

liujoshua: I'm not sure if there's a good way to conditionally include the dependency, but the initialization of Stetho is already conditionally compiled.

Using variants of libraries as dependencies gets tricky, so I made a boolean flag in the xml resources

liujoshua: We should probably use ProGuard for BP release, which would be a good way to remove Stetho at build time

dephillipsmichael: This used to be really easy when there was "compile" keyword instead of "implementation" you just wrote "debugCompile" and only debug variants included the library

liujoshua: There's still debugCompile and debugImplementation. The issue is when you include a library, the library produces one variant by default (defaulted to release). It is possible to publish a different variant or publish multiple variants, but seemed complicated when I've tried in the past.

liujoshua: Also, there is a no-op library (for release variant) for stetho for people but I didn't see one for stetho-okhttp3.

dephillipsmichael: Thanks for taking a look at this.  I'll merge it in.  We can always remove at release.

PR #138, opened 2018-02-22
Update OkHttp3 dependencies, add better handling for upload of videos

liujoshua:


android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
```diff
@@ -149,7 +147,8 @@ public Completable processUploadFiles() {
                                 processUploadForCachedSession(
                                         uploadFile,
                                         uploadDAO.getUploadSession(uploadFile.filename)
-                                )));
+                                ))).doOnError(t -> LOG.warn("Failed to process upload files"))
+                .onErrorComplete();
```
liujoshua, line 156: This will prevent those upload failures from crashing the app. Also, those crashes weren't showing in logs at all, but the added logging now identifies the crash and makes it easier to follow the flow.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
```diff
@@ -77,8 +75,8 @@
     private final OkHttpClient okHttpClient = new OkHttpClient.Builder()
             .connectTimeout(30, TimeUnit.SECONDS)
             .readTimeout(30, TimeUnit.SECONDS)
-            .writeTimeout(30, TimeUnit.SECONDS)
-            .retryOnConnectionFailure(false).build();
+            .writeTimeout(15, TimeUnit.MINUTES)
```
dephillipsmichael: I'd prefer this defaults to 30 seconds as it was, and then in CRF, you set it to be 15 minutes.

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/UploadManager.java
dephillipsmichael: I hate that the crash doesn't show up in logs, it's so confusing

dephillipsmichael (changes requested):
default to 30, change to 15 min in CRF

PR #144, opened 2018-03-06
Bugfix/format of sdk and app versions

liujoshua:


android-sdk/src/main/java/org/sagebionetworks/bridge/android/BridgeConfig.java
```diff
@@ -161,7 +161,7 @@ public String getAcceptLanguage() {
     }
 
     public int getSdkVersion() {
-        return applicationContext.getResources().getInteger(R.integer.osb_android_sdk_version);
+        return BuildConfig.VERSION_CODE;
```
dephillipsmichael, line 164: Hmm didn't you switch this to be the context method because it wasn't getting the correct app's version code?

liujoshua:
Yes I did. That change was for app version. We also want the SDK version
(it's used in the user-agent for HTTP)

dephillipsmichael:
Ah ok gotcha

dephillipsmichael (approved):

PR #145, opened 2018-03-07
Update ResearchStack

liujoshua:


liujoshua:
I think one of our recent PR merges moved RS to an older commit. This RS commit contains a new method which is used by master of one of our apps

PR #146, opened 2018-03-12
Update rest-client

liujoshua:
* Merge of reauth token in UserSessionInfoProvider so reauth isn't lost
* Reset of retry count so retries aren't missed
* Update to how ClientInfo strings are generated

### DIFF
--- a/android-sdk/build.gradle
+++ b/android-sdk/build.gradle
@@ -34,7 +34,7 @@ android {
 }
 
 dependencies {
-    api 'org.sagebionetworks.bridge:rest-client:0.15.1', {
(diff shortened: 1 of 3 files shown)


dephillipsmichael:
I can't see what the code is for UserSessionInfoProvider.mergeReauthToken so I can't really review.  Is there someone better that could?

dephillipsmichael (approved):

liujoshua:
Yea, that was code reviewed. This was refactor moving the code to mergeReauthToken to be reused in BridgeJavaSDK. I left the unit tests covering this in AccountDAO so we're good

PR #154, opened 2018-05-02
Add upload retry to BridgeDataProvider

liujoshua:


dephillipsmichael:
Is there an appropriate place to call this method so that this will happen automatically?  It would be nice if the app dev didn't have to call this method at a specific place to enable this feature.

dephillipsmichael (approved):

PR #171, opened 2018-08-21
Feature/update rest client to reduce timeouts

liujoshua:
* Checked out 73a632a1b88f74add37ed1, which is the base of this branch
* Cherry-picked this commit: "Update rest-client to 0.15.24 and fix compile issues" which addresses API changes since 0.15.6
* Updated the rest-client version
* Merged in head of develop, but will use commit cad05bbfor upcoming release

This is the diff for 73a632a to cad05bb
https://github.com/Sage-Bionetworks/BridgeAndroidSDK/pull/171/files/cad05bb9dbaf3e731c79bb792b70701da65c2ddc
This is the change in payload targeted for the app


liujoshua:
@Robert-Kolmos this may be a good one for us to talk about to go deeper into submodules

PR #172, opened 2018-08-28
Add ScheduledActivity Room DAO

liujoshua:


liujoshua:
Ran into Dex issues when using android library (instead of app) plugin which I couldn't fix, so the instrumentation tests are still in mPower-2.

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/repo/ScheduledActivityRepository.java
```diff
@@ -0,0 +1,4 @@
+package org.sagebionetworks.research.sageresearch.repo;
+
+public interface ScheduledActivityRepository {
```
dephillipsmichael, line 3: What is this class for?

sageresearch-app-sdk/src/androidTest/resources/test_activities_timezones_est.json
```diff
@@ -0,0 +1,32 @@
+{
```
dephillipsmichael, line 1: Should these json classes be here if the tests are still in the other library?

PR #173, opened 2018-08-30
Public API for AppConfig Manager

dephillipsmichael:
Switched app config over to using public api.


DwayneJengSage (approved):
Looks like a straightforward swap from the authentication manager to the public API. Looks good to me!

PR #176, opened 2018-09-11
ScheduleViewModel changes to support study burst

dephillipsmichael:
This PR includes some changes from last PR's feedback.  Also it adds some queries and removes the RsdIdentifier class.

### DIFF
--- a/sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/EntityTypeConverters.kt
+++ b/sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/EntityTypeConverters.kt
@@ -68,7 +68,7 @@ import java.util.UUID
  */
 class EntityTypeConverters {
 
-    private val bridgeGson = GsonBuilder()
+    val bridgeGson = GsonBuilder()
             .registerTypeAdapter(ByteArray::class.java, ByteArrayToBase64TypeAdapter())
             .registerTypeAdapter(LocalDate::class.java, LocalDateTypeAdapter())
             .registerTypeAdapter(DateTime::class.java, DateTimeTypeAdapter())
--- a/sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/EntityTypeConverters.kt
+++ b/sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/EntityTypeConverters.kt
@@ -2,6 +2,11 @@ package org.sagebionetworks.research.sageresearch.dao.room
 
 import android.arch.persistence.room.TypeConverter
 import com.google.gson.GsonBuilder
+import com.google.gson.JsonDeserializationContext
+import com.google.gson.JsonDeserializer
+import com.google.gson.JsonElement
+import com.google.gson.JsonParseException
+import com.google.gson.JsonPrimitive
 import com.google.gson.TypeAdapter
 
 import com.google.gson.reflect.TypeToken
@@ -23,6 +28,7 @@ import org.threeten.bp.Instant
 import org.threeten.bp.LocalDateTime
 import org.threeten.bp.format.DateTimeFormatter
 import java.io.IOException
+import java.lang.reflect.Type
 import java.util.UUID
 
 //
--- a/sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/extensions/DateExtensions.kt
+++ b/sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/extensions/DateExtensions.kt
@@ -41,6 +41,10 @@ fun LocalDateTime.startOfDay(): LocalDateTime {
             .withNano(0)
 }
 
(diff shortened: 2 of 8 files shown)


sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/EntityTypeConverters.kt
```diff
@@ -62,7 +68,7 @@ import java.util.UUID
  */
 class EntityTypeConverters {
 
-    private val bridgeGson = GsonBuilder()
```
liujoshua, line 65: This probably should extend RestUtils.GSON. Could you add a TODO?

RestUtils should expose the GsonBuilder so we can add to it

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/extensions/DateExtensions.kt
```diff
@@ -47,16 +49,40 @@ fun LocalDateTime.startOfNextDay(): LocalDateTime {
             .startOfDay()
 }
 
+fun LocalDateTime.inSameDayAs(date: LocalDateTime): Boolean {
+    return this.isBetweenInclusive(date.startOfDay(), date.endOfDay())
+}
+
+fun LocalDateTime.isBetweenInclusive(start: LocalDateTime, end: LocalDateTime): Boolean {
+    return this == start || this == end ||
```
liujoshua, line 57: should these be this.equals(start) and this.equals(end)?

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/extensions/DateExtensions.kt
```diff
@@ -47,16 +49,40 @@ fun LocalDateTime.startOfNextDay(): LocalDateTime {
             .startOfDay()
 }
 
+fun LocalDateTime.inSameDayAs(date: LocalDateTime): Boolean {
+    return this.isBetweenInclusive(date.startOfDay(), date.endOfDay())
```
liujoshua, line 53: I usually see time intervals implemented as between start (inclusive) and end (exclusive). This would avoid the minusNanos workaround too.

https://docs.oracle.com/javase/8/docs/api/java/time/temporal/ChronoUnit.html#between-java.time.temporal.Temporal-java.time.temporal.Temporal-
https://www.threeten.org/threeten-extra/apidocs/org/threeten/extra/Interval.html

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/extensions/DateExtensions.kt
```diff
@@ -47,16 +49,40 @@ fun LocalDateTime.startOfNextDay(): LocalDateTime {
...
 }
 
 fun ZonedDateTime.startOfNextDay(): ZonedDateTime {
     return this
             .plusDays(1)
             .startOfDay()
+}
+
+fun Instant.inSameDayAs(date: LocalDateTime, timezone: ZoneId): Boolean {
+    return this.isBetweenInclusive(date.startOfDay(), date.endOfDay(), timezone)
```
liujoshua, line 80: I probably would have compared LocalDateTimes, but can't think of why this wouldn't also work

liujoshua (approved):
Minor quibbles

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/EntityTypeConverters.kt
dephillipsmichael: I agree, but RestUtils file is not editable.

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/extensions/DateExtensions.kt
dephillipsmichael: In Kotlin, everything is an object, even primitives, so you don't have to call equals ever.

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/extensions/DateExtensions.kt
dephillipsmichael: Nice, and I do agree with you.  But, I did it this way because MySQL's BETWEEN statement is inclusive for both start and end, and since these extensions work in conjunction with schedule queries, I thought it was best to keep parity with that.

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/extensions/DateExtensions.kt
dephillipsmichael: I'm still getting used to jumping between all these date formats, but I like how java 8 time classes enforce you to have it represented as one vs the Date class which is basically the wild west.

PR #186, opened 2018-10-16
Changes to support mPower2 ResearchStack Surveys

dephillipsmichael:
Depends on https://github.com/ResearchStack/ResearchStack/pull/485


researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -561,6 +562,14 @@ public void addLocalDataGroup(@NonNull String dataGroup) {
     public List<String> getLocalDataGroups() {
         logger.debug("Called getLocalDataGroups");
 
+        // The user session info data groups field is the most up to date reflection of data groups
+        UserSessionInfo userSessionInfo = bridgeManagerProvider.getAuthenticationManager().getUserSessionInfo();
```
dephillipsmichael: I figured this was a reasonable assumption to make, lmk if I'm wrong, however.  The AccountDAO was unpredictable for getting data groups.

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -729,9 +738,20 @@ public void uploadTaskResult(@NonNull TaskResult taskResult) {
...
+     * @param context can be activity or application
+     * @param taskResult from the user doing the task
+     * @param associatedSchedule linked to the task result, for creating metadata and updating schedule on bridge
+     * @param shouldUpdateAssociatedScheduleOnBridge if true, schedule will be updated,
+     *                                               if false, it will only be used to create metadata
+     */
     @SuppressLint("RxLeakedSubscription")    // upload should run as long as it needs to, no early unsubscribe
-    @Override
-    public void uploadTaskResult(Context context, @NonNull TaskResult taskResult) {
+    public void uploadTaskResult(
```
dephillipsmichael, line 752: These parameters allow me to upload to s3 with proper metadata file without having to also update the activity on bridge (which we now do with Room and ScheduleRepository)

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/TaskHelper.java
```diff
@@ -201,13 +202,16 @@ private void cacheSurveyGuidCreatedOn(@NonNull TaskModel taskModel) {
...
-                return surveyFactory.createSmartSurveyTask(context, taskModel);
-            } else {
-                return null;
-            }
-        });
+        return taskModelSingle.map(taskModel ->
+                createSmartSurveyTask(context, taskModel));
+    }
+
+    protected Task createSmartSurveyTask(Context context, @Nullable TaskModel taskModel) {
```
dephillipsmichael, line 209: Exposed to make sub-classing easier

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
liujoshua: Could you clarify? `accountManager.getUserSessionInfo()` delegates to `accountDAO.getUserSessionInfo()` so it seems the predictability should be the same.

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
liujoshua: Since Room and ScheduleRepository for SageResearch tasks and BridgeDataProvider for ResearchStack tasks, don't we always need to update the activity when calling this code path?

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/EntityTypeConverters.kt
```diff
@@ -177,6 +201,77 @@ class EntityTypeConverters {
...
+            scheduleJson = "$scheduleJson \"$s\" : \"${immutableFieldMap[s]}\""
+            if (i < (immutableFieldMap.keys.size - 1)) {
+                scheduleJson = "$scheduleJson,"
+            }
+        }
+        scheduleJson = "$scheduleJson}"
+        try {
+            schedule = RestUtils.GSON.fromJson(scheduleJson, ScheduledActivity::class.java)
+        } catch (e: JsonSyntaxException) {
+            e.printStackTrace()
```
liujoshua: Log?

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/EntityTypeConverters.kt
```diff
@@ -177,6 +201,77 @@ class EntityTypeConverters {
...
+
+    scheduledOn?.let {
+        // TODO: mdephillips 10/14/18 better way to convert from 3tenbp LocalDateTime to jodatime DateTime?
+        val dateTime = DateTime.now()
+                .withYear(it.year)
+                .withDayOfYear(it.dayOfYear)
+                .withHourOfDay(it.hour)
+                .withMinuteOfHour(it.minute)
+                .withSecondOfMinute(it.second)
+        val dateTimeStr = ISODateTimeFormat.dateTime().withOffsetParsed().print(dateTime)
```
liujoshua: Might be better to use RestUtils.GSON.toJson so the same serializer/deserializer is always used

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/EntityTypeConverters.kt
```diff
@@ -177,6 +201,77 @@ class EntityTypeConverters {
...
+ * that the ArchiveUtil.createMetaDataFile() function is concerned with.
+ * @return a ScheduledActivity that can be pass to the BridgeDataProvider and have the metadata JSON included.
+ */
+fun ScheduledActivityEntity.bridgeMetadataCopy(): ScheduledActivity {
+    var schedule = ScheduledActivity()
+    val immutableFieldMap = HashMap<String, String>()
+
+    scheduledOn?.let {
+        // TODO: mdephillips 10/14/18 better way to convert from 3tenbp LocalDateTime to jodatime DateTime?
+        val dateTime = DateTime.now()
```
liujoshua, line 233: val epochMilli = scheduledOn.toInstant().toEpochMilli()
val dateTime = DateTime(epochMilli, DateTimeZone.forID(scheduledOn.zone))

liujoshua (approved):
LGTM. Some questions and comments

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
dephillipsmichael: You would think that, but accountDao.getDataGroups actually references a completely different shared pref field...

`List<String> rawDataGroupList = getValue(KEY_DATA_GROUPS, STRING_LIST);`

dephillipsmichael: Should I switch accountDao.getDataGroups over to just read the user session?

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
dephillipsmichael: Room and ScheduleRepository are decoupled from both RS tasks and SageResearch tasks.  They are just the source of truth for all ScheduleActivity objects in the app, and they always need to stay synced whether a RS task or a SR task was completed.  Therefore when either is complete, I just call ScheduleRepository.updateSchedule(schedule) and it will make the call to bridge and update the db.

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/EntityTypeConverters.kt
dephillipsmichael: I wasn't sure how to log from an extension function, because it doesn't really have a place to put the logger object.  Maybe I can just use LogExt.e() ?

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/EntityTypeConverters.kt
dephillipsmichael: Not a bad idea, let me look into it

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/EntityTypeConverters.kt
dephillipsmichael: Sweet, let me try this.  My way is pretty ridiculous.

dephillipsmichael: That didn't really compile.  toInstant takes a timezone as a parameter, plus there is no zone for scheduleOn because it is a LocalDateTime

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
liujoshua: That is true and the right way to do. ScheduleRepo is in sageresearch-app-sdk and not android-sdk, so I think this only applies because we're using both SageResearch and ResearchStack. Is that right?

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/EntityTypeConverters.kt
liujoshua: Oops, I thought scheduledOn was a ZonedDateTime. need to first to scheduledOn.atZone(ZoneId.systemDefault())
(We're defaulting in a timezone in the DateTime anyway)

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
liujoshua: OH, missed that this is getLocalDataGroups. That was developed for CRF to let a DataGroupQuestionStep temporarily store a local group. It's used in conjunction with a `shouldPersist` flag to control which step should sync to Bridge.

accountDao.getDataGroups perhaps should be moved out of DAO which is geared towards long-term persistence. I think storing in memory in BridgeDataProvider would be a lot less confusing and I'm not sure there's a benefit to saving across application runs

dephillipsmichael: Oh! Hmmm.  Maybe for this particular problem I'll get the user session from BridgeDataProvider instead and read the data groups.

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
dephillipsmichael: Yeah, exactly.

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/EntityTypeConverters.kt
dephillipsmichael: I'm fairly certain scheduledOn.atZone(ZoneId.systemDefault()) will move the time 5 hours (or whatever it is) no?

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/EntityTypeConverters.kt
liujoshua: Ah. We could also make a static Logger on ScheduledActivityEntity or just do LoggerFactory.getLogger(ScheduledActivityEntity.class).warn(msg) inline

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/EntityTypeConverters.kt
liujoshua: I think scheduledOn would be 2018/10/16 08:00 (no zone, since it's a LocalDateTime), calling azZone(ZoneId.systemDefault()) turns it into a ZonedDateTime 2018/10/16 08:00-7:00, and the DateTime should also end up as 2018/10/16 08:00-7:00

dephillipsmichael: oh! perfect. that is what I want

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/EntityTypeConverters.kt
```diff
@@ -177,6 +201,77 @@ class EntityTypeConverters {
...
+
+    schedulePlanGuid?.let {
+        immutableFieldMap.put("schedulePlanGuid", it)
+    }
+
+    if (immutableFieldMap.isNotEmpty()) {
+        // TODO: mdephillips 10/14/18 sync with josh or dwayne to make scheduledOn and schedulePlanGuid mutable
+        // TODO: mdephillips 10/14/18 so I don't have to do this json non-sense
+        var scheduleJson = "{"
+        immutableFieldMap.keys.forEachIndexed { i, s ->
```
liujoshua, line 251: For for the update activity API, the only fields needed/read by the server are guid, startedOn and finishedOn. So when converting ScheduledActivityEntity to ScheduledActivity for a REST call, we don't need to do this. If we're passing other fields to another API/method maybe that API/method should accept a ScheduledActivityEntity or something

dephillipsmichael: That's the clientWritableCopy() function, in which you are right, except for it also needs clientData.  Your comment seems to be about the metadata.json file related ScheduledActivity, which needs all this info.  I based it on ArchiveUtil.createMetaDataFile()

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/EntityTypeConverters.kt
dephillipsmichael: I just unit tested this, and it does indeed shift the time by 4 hours (in my case of EST)

dephillipsmichael: Here is my code, did I do something wrong?

```
val epochMillis = it.atZone(ZoneId.systemDefault()).toInstant().toEpochMilli()
val dateTime = DateTime(epochMillis, DateTimeZone.getDefault())
```

syoung-smallwisdom (approved):
Looks good? I am not seeing anything problematic.

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/viewmodel/SingleLiveEvent.kt
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * A lifecycle-aware observable that sends only new updates after subscription.
+ *
+ * This avoids a common problem with events: on configuration change (like rotation) an update
+ * can be emitted if the observer is active. This LiveData only calls the observable if there's an
+ * explicit call to setValue() or call().
+ *
+ * Note that only one observer is going to be notified of changes.
+ */
+class SingleLiveEvent<T> : MutableLiveData<T>() {
```
dephillipsmichael, line 70: @liujoshua this is a pretty cool class I found online.  I'm using it to return a single event for loading surveys.

liujoshua: I've come across this before. Is this for showing an error message when loading a survey?

There's a lot of discussion how to handle a value that should be only handled once (like displaying a toast or snackbar)
https://github.com/tmurakami/aackt/blob/master/lifecycle-livedata/src/main/java/com/github/tmurakami/aackt/lifecycle/LiveData.kt#L55
https://github.com/googlesamples/android-architecture-components/issues/63

dephillipsmichael: No, actually I added this for the success call for loading a survey from bridge.  I kept the error messaging the same as you made it.  I did notice that when there is an error message, it ALWAYS pops up until you successfully do an action that set it to null, which is annoying.

dephillipsmichael: It's easy to test by just setting your device in airplane mode.  You might want to move the error message over to SingleLiveEvent, but I feel like there probably is a scenario where you would want a sticky error message.

dephillipsmichael:
@liujoshua fixed tests and ready to merge

PR #189, opened 2018-10-20
Survey uploads

dephillipsmichael:
Reverted my BridgeDataProvider in favor of referencing the upload manager directly per our other approaches.  I also moved ResearchStack TaskResult archiving functionality to its on class and var which makes it easy to add custom functionality in mPower, like what I need to do for the StudyBurst task.

### DIFF
--- a/sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/viewmodel/ResearchStackUploadArchiveFactory.kt
+++ b/sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/viewmodel/ResearchStackUploadArchiveFactory.kt
@@ -1,0 +1,172 @@
+/*
+ * BSD 3-Clause License
+ *
+ * Copyright 2018  Sage Bionetworks. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ * 1.  Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * 2.  Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation and/or
+ * other materials provided with the distribution.
+ *
+ * 3.  Neither the name of the copyright holder(s) nor the names of any contributors
+ * may be used to endorse or promote products derived from this software without
+ * specific prior written permission. No license is granted to the trademarks of
+ * the copyright holders even if such marks are included in this software.
+ *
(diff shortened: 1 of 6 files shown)


sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/viewmodel/ScheduleViewModel.kt
```diff
@@ -119,6 +120,25 @@ abstract class ScheduleViewModel(private var scheduleDao: ScheduledActivityEntit
                 }))
     }
 
+    /**
+     * This should only be called with results of old research stack results
+     * @param schedule to make the metadata info json archive, if null, none will be included
+     * @param taskResult to upload to S3
+     */
+    fun uploadResearchStackTaskResultToS3(schedule: ScheduledActivityEntity?,
```
liujoshua, line 128: The ViewModel lifecycle may end before the subscription is complete. It'd be safer to have the ScheduleRepository subscribe (or have both ScheduleRepository and this ViewModel subscribe, calling `.cache()` after uploadManager.queueUpload()` so queueUpload only gets called once).

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/viewmodel/ResearchStackUploadArchiveFactory.kt
```diff
@@ -0,0 +1,172 @@
...
+import org.sagebionetworks.research.sageresearch.dao.room.ScheduledActivityEntity
+import org.sagebionetworks.research.sageresearch.dao.room.bridgeMetadataCopy
+import org.slf4j.LoggerFactory
+import java.util.UUID
+
+/**
+ * The ResearchStackUploadArchiveFactory controls upload archive creation and allows for
+ * easy sub-classing to change the functionality of the factory.
+ */
+open class ResearchStackUploadArchiveFactory: ArchiveFileFactory() {
```
liujoshua, line 55: This class should be in `researchstack-sdk`. Not sure how a `researchstack-sdk` dependency got added to `sageresearch-app-sdk` but we should be keeping them separate. Can you add a TODO?

liujoshua (changes requested):

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/viewmodel/ScheduleViewModel.kt
dephillipsmichael: That's a good point!

PR #190, opened 2018-10-23
Base class functionality for reminder manager and receiver

dephillipsmichael:


sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/reminders/ReminderAlarmReceiver.kt
```diff
@@ -0,0 +1,207 @@
...
+     * @param code associated with this Activity to run when the notification is tapped
+     * @param action of the notification
+     * @return the pending intent which is run when the user taps the notification
+     */
+    protected open fun pendingIntent(context: Context, code: Int, action: String): PendingIntent {
+        val notificationIntent = Intent(context, MainActivity::class.java)
+        notificationIntent.action = action
+        notificationIntent.flags = Intent.FLAG_ACTIVITY_CLEAR_TOP
+        val stackBuilder = TaskStackBuilder.create(context)
+        stackBuilder.addParentStack(MainActivity::class.java)
```
liujoshua, line 124: Do we need to build flexibility around this for non ResearchStack tasks? Or does the task stack not really matter?

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/reminders/ReminderAlarmReceiver.kt
```diff
@@ -0,0 +1,207 @@
...
+     */
+    protected open val notificationChannelId: String
+        get() {
+            return "Sage-Research NotificationChannelId"
+        }
+
+    /**
+     * @property notificationChannelId set on Android OS >= 26, explains your notifications
+     */
+    protected open val notificationChannelTitle: String
```
liujoshua: Nit: make channel title/desc a string resource since these are displayed to the user

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/reminders/ReminderManager.kt
```diff
@@ -0,0 +1,251 @@
...
+    val guid: String,
+    /**
+     * @property action notification action
+     */
+    val action: String,
+    /**
+     * @property code notification code
+     */
+    val code: Int,
+    /**
```
liujoshua, line 247: Thinking about reminders and time zone changes... I think we should schedule/save reminders as a LocalTime (or LocalDateTime) and respond to time zone changes to cancel/reschedule the alarmEpochMillis. What do you think?

Looks like you have this filter but I didn't see how we are handling time zone
```
        <intent-filter>
            <action android:name="android.intent.action.TIMEZONE_CHANGED " />
        </intent-filter>
```

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/reminders/ReminderManager.kt
```diff
@@ -0,0 +1,251 @@
...
+        intent.data = Uri.parse(reminder.guid)
+        return PendingIntent.getBroadcast(context, reminder.code, intent, PendingIntent.FLAG_ONE_SHOT)
+    }
+
+    /**
+     * @param context can be app or activity
+     * @return a shared preferences object used to store PendingIntents
+     */
+    protected fun pendingIntentPrefs(context: Context): SharedPreferences {
+        return context.getSharedPreferences("ReminderManagerPendingIntents", MODE_PRIVATE)
```
liujoshua: Can we make this a static val?

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/reminders/ReminderAlarmReceiver.kt
dephillipsmichael: Good point!

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/reminders/ReminderAlarmReceiver.kt
dephillipsmichael: The task stack does matter.  There is flexibility already built in because this function is open to sub-classes.

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/reminders/ReminderManager.kt
dephillipsmichael: I'd rather not.  There's no reason it should stick around for the lifetime of the app.

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/reminders/ReminderManager.kt
dephillipsmichael: This is a good point.  If the timezone changed, the epochMillis would no longer be at the hour/minute that you scheduled it at.  I'll switch over to using LocalDateTime instead, and I will compute the epochMillis before scheduling, based on the user's current timezone.

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/reminders/ReminderManager.kt
dephillipsmichael: However, I should make it a member variable.

liujoshua (approved):

PR #192, opened 2018-11-07
SBAReportManager equivalent ReportRepository

dephillipsmichael:
Create the ReportRepository with the ReportEntityDao with corresponding unit tests.

### DIFF
--- a/sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/ReportEntity.kt
+++ b/sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/ReportEntity.kt
@@ -78,13 +78,3 @@ data class ReportEntity(
      */
     @ColumnInfo(index = true)
     var needsSyncedToBridge: Boolean? = null)
-
-/**
- * The category for a given report. This is used to determine whether the `dateTime` or `localDate`
- * properties on `ReportDataEntity` are used to group the reports.
- */
-enum class ReportCategory(val rawValue: String) {
-    SINGLETON("singleton"),
-    GROUP_BY_DAY("groupByDay"),
-    TIMESTAMP("timestamp")
-}
--- a/sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/ReportEntity.kt
+++ b/sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/ReportEntity.kt
@@ -1,0 +1,90 @@
+/*
+ * BSD 3-Clause License
+ *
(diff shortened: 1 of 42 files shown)


syoung-smallwisdom (approved):
looks good

dephillipsmichael:
Thanks @syoung-smallwisdom !  I'm going to wait for @liujoshua to have a look too before merging

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/ReportEntity.kt
```diff
@@ -0,0 +1,80 @@
...
+     * @property data open ended formatted data contained within the report
+     */
+    @SerializedName("data")
+    var data: ClientData? = null,
+    /**
+     * @property dateTime of when the report was completed when ReportCategory is singleton or timestamp
+     */
+    @SerializedName("dateTime")
+    @ColumnInfo(index = true)
+    var dateTime: Instant? = null,
```
dephillipsmichael, line 67: @syoung-smallwisdom wanted to verify this with you... I didn't even end up reading/writing the "date" field anywhere in the db, because it always seemed redundant with the real value being stored in either localDate or dateTime.  Is that a safe assumption?

liujoshua (approved):

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/ReportEntity.kt
syoung-smallwisdom: Yes, that is a valid assumption. I suspect Erin handled it the way she did b/c there isn't an iOS equivalent to `Instant` verse `LocalDate` so she created the `date` property but that isn't how you set the date or how the server stores the date.

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/ScheduleRepository.kt
```diff
@@ -155,33 +157,63 @@ open class ScheduleRepository constructor(
...
+    /**
+     * @property asyncScheduler for performing network and database operations
+     */
+    @VisibleForTesting
+    protected open val asyncScheduler: Scheduler get() = Schedulers.io()
+
+    /**
+     * @property scheduleRepoErrorLiveData for monitoring error messages from the schedule repository
+     */
+    val scheduleRepoErrorLiveData = MutableLiveData<String>()
```
dephillipsmichael, line 169: @liujoshua I moved the error live data to the schedule repo so that the repository could be the one managing the subscriptions while still be able to communicate back to the view layer.

PR #197, opened 2018-11-25
AA-398

dephillipsmichael:
Fixed metadata archive, and added answers map json archive to uploads


sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch_app_sdk/archive/AbstractResultListArchiveFactory.java
```diff
@@ -0,0 +1,55 @@
...
+import com.google.common.collect.ImmutableSet;
+
+import org.sagebionetworks.bridge.data.ArchiveFile;
+import org.sagebionetworks.research.domain.result.interfaces.Result;
+
+/**
+ * The AbstractResultListArchiveFactory enables an archive factory to group multiple results
+ * into a single archive file.
+ */
+public interface AbstractResultListArchiveFactory extends AbstractResultArchiveFactory {
```
liujoshua: Seems like AbstractResultListArchiveFactory and AbstractResultListArchiveFactory.ResultListArchiveFactory don't need to extend from ActractResultArchiveFactory and ResultArchiveFactory

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch_app_sdk/archive/SageResearchResultArchiveFactory.java
```diff
@@ -38,4 +39,23 @@ public SageResearchResultArchiveFactory(ImmutableList<ResultArchiveFactory> arch
         LOGGER.warn("Result archive factory not found for result: {}, skipping", result);
         return ImmutableSet.of();
     }
+
+    @NonNull
+    @Override
+    public ImmutableSet<? extends ArchiveFile> toArchiveFiles(@NonNull final ImmutableList<Result> resultList) {
+        ImmutableSet.Builder<ArchiveFile> builder = new ImmutableSet.Builder<>();
+        for (ResultArchiveFactory archiveFactory: archiveFactories) {
+            // Check both interface options for archives
```
liujoshua: We shouldn't need to check both types. The AbstractFactory pattern is that the AbstractFactory uses Factories, so in here we'd call the AbstractResultListArchiveFactory, which in turn knows about all the types of ResultListArchiveFactories the app supports

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch_app_sdk/archive/TaskResultArchiveFactory.java
```diff
@@ -26,12 +34,36 @@ public boolean isSupported(@NonNull final Result result) {
...
-        for (Result asynResult : ((TaskResult) result).getAsyncResults()) {
+
+        // Create the archives for the async results
+        List<Result> asyncResults = taskResult.getAsyncResults();
+        for (Result asynResult : asyncResults) {
             builder.add(abstractResultArchiveFactory.toArchiveFiles(asynResult).toArray(new ArchiveFile[]{}));
         }
+
+        // Create the archives for the result list archive factories
+        if (abstractResultArchiveFactory instanceof AbstractResultListArchiveFactory) {
```
liujoshua: Having toArchiveFiles(@NonNull final ImmutableList<Result> resultList) implemented by SageResearchResultArchiveFactory muddles the AbstractResultArchiveFactory pattern. I'm thinking having TaskResultArchiveFactory use AnswerMapResultListFactory (or ResultListFactory) would be clearer.

liujoshua (commented):
Great work. Some design comments

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch_app_sdk/archive/TaskResultArchiveFactory.java
dephillipsmichael: I agree with all your comments.  I switched over to the SageResearchResultArchiveFactory to take in an injected TaskResultArchiveFactory that the app can choose to provide the base one or a sub-class I made called TaskResultAnswerMapArchiveFactory that adds the answer map json archive along with the same functionality it had before.  Check out the code now and lmk if you think this is better.

Also, I think this will create a problem when we move to multiple nested TaskResults. I think each task result will create their own "answers" json archive and only the last one in the list will be upload, we may need some sort of archive merging functionality when we start doing that.

liujoshua (approved):

PR #198, opened 2018-11-26
AA-303

dephillipsmichael:
Added support for custom result reports.

### DIFF
--- a/sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/ReportResultDataMap.kt
+++ b/sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/ReportResultDataMap.kt
@@ -1,0 +1,111 @@
+/*
+ * BSD 3-Clause License
+ *
+ * Copyright 2018  Sage Bionetworks. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ * 1.  Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * 2.  Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation and/or
+ * other materials provided with the distribution.
+ *
+ * 3.  Neither the name of the copyright holder(s) nor the names of any contributors
+ * may be used to endorse or promote products derived from this software without
+ * specific prior written permission. No license is granted to the trademarks of
+ * the copyright holders even if such marks are included in this software.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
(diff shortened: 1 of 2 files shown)


sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/extensions/ResultExtensions.kt
```diff
@@ -74,9 +75,15 @@ fun org.researchstack.backbone.result.TaskResult.clientDataAnswerMap(): Map< Str
 fun TaskResult.clientDataAnswerMap(): Map<String, Any> {
     val answerMap = TreeMap<String, Any>()
     this.flattenResult().forEach { result ->
-        (result as? AnswerResult<*>)?.let { answerResult ->
-            answerResult.answer?.let {
-                answerMap.put(answerResult.identifier, it)
+        (result as? ReportResultDataMap)?.let { customDataMap ->
```
liujoshua, line 89: Does this mean ReportResultDataMap is a Result?

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/ReportResultDataMap.kt
```diff
@@ -0,0 +1,111 @@
...
+class ReportResultDataMapHelper {
+    companion object {
+        /**
+         * @param src any object that can be represented and merged into a ReportEntity.
+         * @return a data map identical to how this object looks as json,
+         *         that will properly merge with other TaskResult's stepHistory data maps in a ReportEntity.
+         */
+        @JvmStatic fun toDataMap(src: Any): Map<String, Any>? {
+            val bridgeGson = EntityTypeConverters().bridgeGson
+            val json = bridgeGson.toJsonTree(src).asJsonObject
```
liujoshua, line 103: We should call isJsonObject if we're taking in an Any

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/extensions/ResultExtensions.kt
```diff
@@ -74,9 +75,15 @@ fun org.researchstack.backbone.result.TaskResult.clientDataAnswerMap(): Map< Str
...
+            customDataMap.toDataMap()?.forEach {
+                answerMap[it.key] = it.value
+            }
+        } ?: run {
```
liujoshua, line 93: Could this have been used to create the answer.json for archiving instead of ResultListArchiveFactory?

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/ReportResultDataMap.kt
```diff
@@ -0,0 +1,111 @@
...
+
+package org.sagebionetworks.research.sageresearch.dao.room
+
+import com.google.gson.Gson
+import java.util.TreeMap
+
+/**
+ * Implementing ReportResultDataMap allows for Result objects to provide custom Report data maps.
+ */
+interface ReportResultDataMap {
```
liujoshua, line 41: I'm not sure about how this will be used. Seems like a report would be tied to a known type so we would know reportId = "report1" corresponds to Report1Model.class

liujoshua (approved):
Looks good? Some questions

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/extensions/ResultExtensions.kt
dephillipsmichael: Good point!  I updated the archive factory branch to use this instead.  It also fixed the nested TaskResult bug by including only the top-level TaskResult answers map, which will traverse the nested TaskResults and include their answer maps as well.

PR #208, opened 2018-12-04
Bugfix/fix broken rx chain for upload

liujoshua:


sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch_app_sdk/TaskResultUploader.java
```diff
@@ -103,65 +102,85 @@ public TaskResultUploader(@NonNull final BridgeConfig bridgeConfig,
...
-
-        for (ArchiveFile resultArchiveFile : abstractResultArchiveFactory.toArchiveFiles(taskResult)) {
-            builder.addDataFile(resultArchiveFile);
-        }
+        String archiveFilename = schemaKey.getId() + schemaKey.getRevision() + taskResult.getTaskUUID();
 
-        String archiveFilename = sk.getId() + sk.getRevision() + taskResult.getTaskUUID();
-
-        Completable uploadToS3Completable = Completable.fromAction(() ->
-                uploadManager.queueUpload(archiveFilename, builder.build())
```
liujoshua, line 128: Hard to see bug. The `Action` gets called but the `rx.Completable` it produces is ignored (never attached to the rest of the Rx chain).

dephillipsmichael: I don't understand, why wouldn't this code... 

```
return metadataCompletable(builder, taskResult)
                 .concatWith(uploadToS3Completable);
```

Not concatenate the upload to S3 action -> flatMapCompletable (upload it to s3).  What am I missing about RX here?

PR #218, opened 2019-03-06
consent_bug

dephillipsmichael:
Propagate consent required exception if we don't have it saved locally.

### DIFF
--- a/android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
+++ b/android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
@@ -533,13 +533,15 @@ public class AuthenticationManager implements UserSessionInfoProvider.UserSessio
                             ConsentStatus consentStatus = consentStatusEntry.getValue();
 
                             // required consent missing on Bridge and present locally
-                            if (consentStatus.isRequired()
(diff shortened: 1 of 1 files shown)


android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
```diff
@@ -533,13 +533,15 @@ public Completable requestPhoneSignIn(@NonNull String regionCode, @NonNull Strin
...
-                                // upload local consents, fail this sign in if consent upload fails
-                                return uploadLocalConsents()
-                                        .toSingle();
+                            if (consentStatus.isRequired() && !consentStatus.isConsented()) {
+                                if (isConsentedInLocal(subpopulationGuid)) {
+                                    // upload local consents, fail this sign in if consent upload fails
+                                    return uploadLocalConsents().toSingle();
+                                } else {
+                                    // No saved consents for the required guid, propagate the error
+                                    return Single.error(new ConsentRequiredException
```
liujoshua, line 547: Alternately, this code could be moved to  onErrorResumeNext and you can just propagate the original exception

liujoshua (approved):
Looks good, can you update the javadoc to say it will throw exception when

PR #222, opened 2019-04-26
Session token fixes

dephillipsmichael:
Right now, after a user successfully signs up and clicks on their auth link in their email, they see an error pop-up saying we could not get their user information. When they return to the app, some users are getting in fine, and others it seems are not.

On Monday, the bridge team pushed code that made the session refresh after the user uploads consent.  While this should ideally not have been an issue for our app, it exposed a few bugs in our code that starting causing reproducible user facing errors on sign up. 

Here are the bugs and their solutions:

1. Session token from signIn call isn’t being saved. This leads to the first 401 when attempting to sign consent.  The bug was because createAuthStateFromStoredCredentials was always overwriting the session token to be a blank sting "", when in most cases, we wanted it to keep its value, like after a successful sign in.
2. Server is switching the session token when you sign consent. This the second 401 when the app calls get activities for the first time.  This was what caused the app to get in one more bad state that pushed it over the edge. 
3. App calls get activities twice, seemingly simultaneously.  This is because when you sign up, we request 3 weeks of activities (14 day request and then a 7 day request) in advance to the app can work offline for the duration of the study.  To solve this, I switched over to a max duration of 30 days, which is something Dwayne informed me they added to Bridge recently.  This makes it so this scenario never happens anymore, and is also a fix in BP repo, not here. https://github.com/Sage-Bionetworks/BloodPressureApp-Android/pull/505

### DIFF
--- a/android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
+++ b/android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/AuthenticationManager.java
@@ -51,6 +51,7 @@ import javax.inject.Inject;
 import rx.Completable;
 import rx.Observable;
 import rx.Single;
+import rx.functions.Action1;
 import rx.functions.Func1;
 
 /**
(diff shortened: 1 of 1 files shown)


dephillipsmichael:
I merged this in to do a build, but it still needs reviewed.

PR #232, opened 2019-10-22
Profile settings

nategbrown9:


android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/models/ProfileDataSource.kt
```diff
@@ -0,0 +1,194 @@
+package org.sagebionetworks.bridge.android.manager.models
+
+/**
+ * A ProfileDatSource is a configuration element defined in the Bridge Study Manager. It is used to
```
liujoshua: typo in ProfileDataSource

sageresearch-app-sdk/src/androidTest/java/org/sagebionetworks/research/sageresearch/dao/room/ResourceEntityDaoTests.kt
```diff
@@ -0,0 +1,62 @@
...
+        Assert.assertEquals(resourceEntity, getValue(resourceDao.getResource(RESOURCE_ID)))
+    }
+
+    @Test
+    fun test_clear() {
+        resourceDao.upsert(resourceEntity)
+        resourceDao.clear()
+        Assert.assertNull(getValue(resourceDao.getResource(RESOURCE_ID)))
+    }
+//    @Rule
```
liujoshua: nit: remove unneeded lines

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/SurveyRepository.kt
```diff
@@ -0,0 +1,44 @@
+package org.sagebionetworks.research.sageresearch.dao.room
+
+import io.reactivex.Completable
+import io.reactivex.Flowable
+import io.reactivex.Single
+import org.joda.time.DateTime
+import org.sagebionetworks.bridge.android.manager.SurveyManager
+import org.sagebionetworks.bridge.rest.model.Survey
```
liujoshua, line 8: I'm a wary of directly using REST models for persistence since these have occasionally changed in breaking ways and we aren't great about version upgrade testing.

That said, I don't think we've broken the serialization contract in a non-backwards-compatible way so we may be okay from a DB integrity perspective, just compile-time breaks, which would be annoying but less pernicious.

sageresearch-app-sdk/src/androidTest/java/org/sagebionetworks/research/sageresearch/dao/room/ResourceEntityDaoTests.kt
liujoshua: Remove

researchstack-sdk/src/main/java/org/sagebionetworks/bridge/researchstack/BridgeDataProvider.java
```diff
@@ -152,15 +152,19 @@ public String getStudyId() {
     //region AppConfig
 
     /** Get app config from the cache, or fall back to server if there is no value in the cache. */
```
liujoshua: Looks like we can remove the entire `//region AppConfig` block

android-sdk/src/main/java/org/sagebionetworks/bridge/android/manager/ParticipantRecordManager.java
```diff
@@ -68,10 +71,17 @@ public StudyParticipant getCachedParticipantRecord() {
      */
     @NonNull
     public Single<StudyParticipant> getParticipantRecord() {
-        return toBodySingle(authStateHolderAtomicReference.get().forConsentedUsersApi
-                .getUsersParticipantRecord(false))
-                .doOnSuccess(accountDAO::setStudyParticipant)
-                .doOnError(throwable -> logger.error(throwable.getLocalizedMessage()));
+        if (lastLoad + DEFAULT_REFRESH > System.currentTimeMillis()) {
```
liujoshua, line 74: We should come up with a standard place for debouncing/caching calls, whether it's in Repository/Manager or elsewhere.

I poked around looking at where this is used and this change seems fine. But I'd find it clearer if `updateParticipantRecord` either 1) always attempts an API call to eagerly refresh the local cache (this is a somewhat odd workaround due to the fact that API returns a `UserSessionInfo` instead of the updated `StudyParticipant`) or 2) does not call `getParticipantRecord` itself and relies on `getParticipantRecord`'s caller to trigger a lazy refresh

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/AppConfigRepository.kt
```diff
@@ -0,0 +1,66 @@
...
+
+class AppConfigRepository(resourceDao: ResourceEntityDao, val appConfigManager: AppConfigManager): ResourceRepository<AppConfig>(resourceDao) {
+
+    companion object {
+        const val APP_CONFIG_ID = "AppConfigId"
+    }
+
+    var cachedAppConfig: AppConfig? = null
+
+    val appConfig : Single<AppConfig>
```
liujoshua: As we talked about, consider exposing `Flowable` at the Repository layer

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/ResourceEntity.kt
```diff
@@ -0,0 +1,51 @@
...
+import java.lang.reflect.Type
+
+@Entity
+data class ResourceEntity(
+
+    /**
+     * @property identifier the resource identifier
+     */
+    @PrimaryKey
+    var identifier: String,
```
liujoshua, line 16: What do you think about making a composite primary key of identifier + type? In some cases our Bridge identifiers aren't GUIDs

liujoshua (approved):

PR #237, opened 2019-12-11
HistoryItemEntity room table to support history view.

nategbrown9:


sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/HistoryItemEntity.kt
```diff
@@ -0,0 +1,70 @@
...
+
+        /**
+         * @property reportId of the report backing this history item
+         */
+        var reportId: String,
+
+        /**
+         * @property dateBucket day string in format: yyyy-MM-dd
+         */
+        var dateBucket: String,
```
liujoshua: Would this be better as a LocalDate?

sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/dao/room/HistoryItemEntity.kt
```diff
@@ -0,0 +1,70 @@
...
+
+        /**
+         * @property dateTime of when the history item occurred
+         */
+        var dateTime: Instant
```
liujoshua: Should we use ZonedDateTime? Depending on what Bridge sends or what iOS does.. we get bitten a lot when we store only UTC time or an OffsetDateTime and lose the Zone info

sageresearch-app-sdk/src/main/res/layout/fragment_profilesettings_list.xml
```diff
@@ -28,7 +28,7 @@
         android:textSize="24sp"
         android:textStyle="bold"
         android:textColor="@color/rsb_white"
-        android:text="Your Profile"/>
+        android:text="@string/your_profile"/>
```
liujoshua, line 31: Nice fix

liujoshua (approved):

liujoshua (approved):

PR #239, opened 2020-06-16
Support for Passive Gait Recording

kalperin:
This pull request has support for displaying the preferences for passive data recording consent as well as updated null handling.  Gradle is used to build SageResearch-Android as a dependency via a local repository.  If SageResearch-Android is published to a maven repo then this grade file can be updated.


sageresearch-app-sdk/src/main/java/org/sagebionetworks/research/sageresearch/profile/ProfileSettingsRecyclerViewAdapter.kt
```diff
@@ -193,10 +203,32 @@ class ReportProfileItemRow(val reportProfileDataItem: ReportProfileDataItem,
                            profileItem: ProfileItemProfileTableItem,
                            profileDataLoader: ProfileDataLoader) : ProfileItemRow(profileItem, profileDataLoader) {
 
+    override val detail: String?
+        get() {
+            var value = profileDataLoader.getValueString(profileItem.profileItemKey)
```
kalperin, line 208: Is there a `profileDataLoader.getValueBool()` that would standardize for different representations of a `bool`.  If not, then let's use `Boolean.parseBool()`

kalperin:
Need to wait for the SageReseach-Android push before we approve, but this looks good.

build.gradle
```diff
@@ -54,6 +54,12 @@ buildscript {
     classpath "org.jetbrains.dokka:dokka-android-gradle-plugin:$dokka_version"
     classpath "com.jakewharton:butterknife-gradle-plugin:$butterknife_version"
 
+
+    classpath "org.sagebionetworks.research:domain:$sage_research_version"
+    classpath "org.sagebionetworks.research:presentation:$sage_research_version"
+    classpath "org.sagebionetworks.research:data:$sage_research_version"
```
nategbrown9: Shouldn't need to add these here. The sub modules that need sage research depend on them directly.

nategbrown9 (changes requested):
Just need to remove the added dependencies in the top level gradle file.

build.gradle
elkasvirElka: Fixed

nategbrown9 (approved):

PR #240, opened 2021-05-07
Refactored sign up function to a more flexible implementation.

dephillipsmichael:
MP2 has the requirement to add data groups when a user signs up with a phone number, but unfortunately, all the available functions in BridgeDataProvider for phone number take the Phone class as a parameter.  This does not allow for any data groups to be attached to the SignUp model.

I'm open to suggestions that fix the issue, but this PR has my first draft solution for this by adding a new function that can do phone sign up and also allow the developer to provide data groups.


dephillipsmichael:
That function is already exposed through the injected AuthenticationManager, so closing this PR.

